Issue #58, opened 2023-10-21
inconsistent validation results for high-precision floating point `output_type_id`s


elray1:
I attach (at bottom) two example submission files and two example `tasks.json` config files adapted from the set up in FluSight. In both files, a quantile level of `0.1` is represented in the file as `0.1000000000000000055511`, which is the floating point representation of that value (at least, on my machine). The difference is whether the hub additionally includes another target with pmf `output_type`, where the `output_type_id` is a character string.

When I run validations for these output files, I get the following results:

# Using the tasks.json file that includes a character output_type:

For the submission file that includes the pmf outputs, I see the following output:

```
> hubValidations::validate_submission(
+     hub_path=".",
+     file_path="UMass-gbq_bootstrap/2023-10-28-UMass-gbq_bootstrap.csv")

...

✖ 2023-10-28-UMass-gbq_bootstrap.csv: `tbl` contains invalid values/value combinations.  Column
  `output_type_id` contains invalid value "0.1000000000000000055511".
```

For the submission file that does not include the pmf outputs, I get a similar result:

```
> hubValidations::validate_submission(
+     hub_path=".",
+     file_path="UMass-gbq_bootstrap/2023-11-04-UMass-gbq_bootstrap.csv")

...

✖ 2023-11-04-UMass-gbq_bootstrap.csv: `tbl` contains invalid values/value combinations.  Column
  `output_type_id` contains invalid value "0.1000000000000000055511".
```

# Using the tasks.json file that does not include a character output type

However, for the submission file that does not include the pmf outputs, I get the following output with no errors related to this check:

```
+     file_path="UMass-gbq_bootstrap/2023-11-04-UMass-gbq_bootstrap.csv")
✔ FluSight-forecast-hub: All hub config files are valid.
✔ 2023-11-04-UMass-gbq_bootstrap.csv: File exists at path
  model-output/UMass-gbq_bootstrap/2023-11-04-UMass-gbq_bootstrap.csv.
✔ 2023-11-04-UMass-gbq_bootstrap.csv: File name "2023-11-04-UMass-gbq_bootstrap.csv" is valid.
✔ 2023-11-04-UMass-gbq_bootstrap.csv: File directory name matches `model_id` metadata in file name.
✔ 2023-11-04-UMass-gbq_bootstrap.csv: `round_id` is valid.
✔ 2023-11-04-UMass-gbq_bootstrap.csv: File is accepted hub format.
[...]

PR #60, opened 2023-11-01
Improve handling of numeric output type IDs.

### LINKED ISSUE #54: Unexpected error about `output_type_id` data type with FluSight submission
Validating the [attached file](https://github.com/Infectious-Disease-Modeling-Hubs/hubValidations/files/12851968/2023-10-14-UMass-gbq_bootstrap.csv)
 yields the following error:
```
! 2023-10-14-UMass-gbq_bootstrap.csv: Column data types do not match hub schema.  `output_type_id ` should be "character " not "double "
```

I don't think this errors makes sense, since that column could be read in with a character type; a data type issue here is more an issue with how the file was read in rather than how the data was encoded.

annakrystalli:
Resolves #54 & #58


R/check_tbl_values.R
```diff
@@ -105,3 +107,40 @@ summarise_invalid_values <- function(valid_tbl, accepted_vals) {
...
+    type_coerce <- tbl[["output_type"]] %in% num_output_types
+    num_output_type_id <- suppressWarnings(
+      as.numeric(tbl$output_type_id[type_coerce])
+    )
+    # establish only valid coercions to distinguish between the potential for
+    # two cdf output types in the same round, one numeric and one character.
+    valid <- !is.na(num_output_type_id)
+    tbl$output_type_id[type_coerce][valid] <- as.character(
+      num_output_type_id[valid]
+    )
```
elray1, line 143: I think the use of `as.character` here would open up the problems I outlined in discussion #59. I added an example of a unit test that passes, but that I think should not pass, [here](https://github.com/Infectious-Disease-Modeling-Hubs/hubValidations/pull/61). In that example, the `output_type_id` in the submission file is `0.09999999999999997779554`. On my system, this is not equal to the floating point representation of 0.1, but `as.character` casts it to `"0.1"`. Here's an illustration:

```
> a <- 0.0999999999999999777955
> b <- 0.1000000000000000055511
> c <- 0.1
> a == b
[1] FALSE
> a == c
[1] FALSE
> b == c
[1] TRUE
> as.character(a) == as.character(b)
[1] TRUE
> as.character(a) == as.character(c)
[1] TRUE
> as.character(b) == as.character(c)
[1] TRUE
> as.character(a)
[1] "0.1"
```

As a result of this, if we use values coerced via `as.character` to do validations, we may be generous in passing submissions that `as.character` rounds to the value specified in the tasks config file. This would then cause issues later on in places like ensembling code or filtering code, where grouping or filtering on the `output_type_id` may give unexpected results.

I am wondering if a way forward might be to do the data type coercion in a different order, or just deal with character representations from the start for purposes of validating `output_type_ids` and task id variables? Something like:
1. For purposes of validation, read in all fields other than the `value` column as character strings. Additionally, read the contents of `tasks.json` as strings if that's possible (may not be?)?
2. Check exact equality of values for `output_type_id`s and task id variables with the required/optional values specified in `tasks.json`, as strings. For example, if the `tasks.json` file says that a valid quantile level is 0.1, we expect to see the exact characters `"0.1"` in the submission file.  This would change our current passing validation results for submissions with values like 0.1000000000000000055511 to failures.

Something like this would entirely sidestep the issues of what gets rounded to 0.1 or what the floating point representation of 0.1 is.

Another option might be to use arrow's `cast` function for type conversion instead of `as.character`, since it seems to be more careful in what it rounds. But my question about that is whether the outputs here would be stable across platforms. For example, maybe on one computer the floating point representation of 0.1 is 0.1000000000000000055511 and on another computer (e.g., one that's not using 64 bit representations of numbers) it's something slightly different. For that reason, it seems better to avoid relying on equality of floating point representations if possible.

elray1 (changes requested):

R/check_tbl_values.R
annakrystalli: Thanks @elray1 . Great points and I agree that although some output IDs are numeric, given we want to be matching output ID values effectively as categories / use them for grouping (both in validation and downstream) it feels  indeed more appropriate to work with data as character for this purposes and match exact values, not check for equality using numeric tolerances or rounding behaviours.

So I will look into the most efficient way of including all character versions of data in our model output validation checks.

I will merge your currently failing test to include in the test suite. I also feel the original test you included (of a `0.1000000000000000055511` value included instead of a `0.1` value should now also fail.

### Persistent issue with trailing zeros in `tasks.json`

The one remaining _problem_ however that `as.character()` was solving for us is removing trailing zeros. When reading the `tasks.json` into R from JSON and then passing values to `expand.grid` to create the grid of all possible valid values, any trailing zeros are removed from numeric output type IDs. Although I spent quite some time trying to figure out a way to read in values from JSON in R as character (retaining any trailing zeros) I could not find a way so I don't think the suggestion:
> Additionally, read the contents of tasks.json as strings if that's possible (may not be?)?
is currently possible.

Using `as.character()` on model output data as proposed in this PR was addressing this issue:
``` r
as.character(c(0.300, 0.350))
#> [1] "0.3"  "0.35"
```
so that including `0.300` as an output type ID in a `tasks.json` would match a `0.300` value in the data (which previously failed because the value from config would be converted to `0.3` while the value in model output data would be read in as `0.300`. Having said that, the approach in this PR is still a bit of a hack, as, to match values, trailing zeros are removed from both which is not quite right.  I think we will have to tackle this with strong documentation and perhaps some sort of validation check (although I think a validation check will be difficult for the same reasons. Detecting trailing zeros in numerics will likely be hampered by the requirement to retain trailing zeros...our original problem!

annakrystalli: OK Just to clarify what the new approach I'm putting together is doing:

- The `read_model_out_file` fn now has argument `coerce_types` with values `hub`, `chr` and `none` instead of logical `use_hub_schema`.
    - `hub` reads in (csvs) or coerces (parquet & arrow which have schema defined in file,  using arrow `cast`) to the hub schema.
    -  `chr` reads in (csvs) or coerces (parquet & arrow) all columns as character.
    - `none`: no coercion. Read using default `arrow::read_*` fun settings.
  
This now causes both `0.1000000000000000055511` and `0.09999999999999997779554` to fail validation. It also means that values with trailing zeros (e.g. `0.300`) will always fail, regardless of whether `0.300` is specified in `tasks.json` (leading to the issue discussed in #54). While I do feel this is actually correct behaviour when validating model output data, it still feels unsatisfactory to me form the aspect that values with trailing zeros can currently be specified in config files but that will not be respected when reading in and will cause validation errors. I think this will remain problematic until we find a way to detect and throw a warning about trailing zeros during config file validation.

annakrystalli: ### Confirm read and column type checking behaviour for different file types

Also can we just confirm that you are happy with the following column type checks (same as before after addressing issues discussed in #54) but needs some recoding due to changed default behaviour in `read_model_out_file`).

- When reading in `csv` files (which do not have column type information embedded), whether a file can be parsed using the hub schema on read is checked via `check_file_read` fn. If that check passes, the file is then read in using the file schema. If a csv is read in therefore, it will always pass the column type check.
- For arrow and parquet files, files are read in according to the file schema (not coerced to the hub schema). As such, the check on column types checks that the file schema matches the hub schema (and may therefore fail).

Could just give me a thumbs up if you agree?

Issue #64, opened 2023-12-06
Add more detailed documentation of what each test in validation functions entails


annakrystalli:
From @LucieContamin :
> I was testing it on some other SMH entries and I am confused by the output, do you have a list of all the checks it's running by default? SO I can see exactly what is expected to be checked?

Fair point!

Need to add a more detailed vignette and docs to the functions regarding the contents of the high level validation functions and point to more details for each check. Also instructions on how to access additional information from check objects (Ties in to #56)

PR #66, opened 2023-12-19
Add finer control of file mods/deletion checks. Resolves #65

### LINKED ISSUE #65: Output error/warning/message when model output updated
Following the issue #44 and the PR #63, I wonder if the test to check for modifications/deletion of a previously submitted model output files should have an additional parameter with the possibility to have the check either:
- returning an error and failing the validation
- OR returning a warning/message and passing the validation

I am thinking for the Scenario projection hubs, it happen sometimes that a team wants to update their submission during an ongoing run and in this case we will allow the team to modify(or even remove) their submission. 

We can technically bypass the failed validation and still merge the PR manually but I think it might be more clear to the team submitting to have a message/warning than an error (in the case we allow modification)

annakrystalli:
Hey @LucieContamin & @elray1 !

I've taken all your comments on board and introduced two new arguments to`validate_pr()` for controlling modification/deletions checks performed on model output and model metadata files.
  - `file_modify_check`, which controls whether modification/deletion checks are performed and what is retuned if modifications/deletions detected. It takes the following options: 
    - `"error"`: Appends a `<error/check_error>` condition class object for each applicable modified/deleted file.
    - `"warning"`: Appends a `<warning/check_warning>` condition class object for each applicable modified/deleted file.
    - `"message"`: Appends a `<message/check_info>` condition class object for each
applicable modified/deleted file.
    - `"none"`: No modification/deletion checks performed.
  - `allow_submit_window_mods`, which controls whether modifications/deletions of model output files are allowed within their submission windows.

They main issue I hit that ended up in some more time and complicated code than I hoped for is summarised as a note in the function docs too:

Note that to establish relative submission windows when performing modification/deletion checks and `allow_submit_window_mods` is `TRUE`, the reference date is taken as the round_id extracted from the file path (i.e. `submit_window_ref_date_from` is always set to `"file_path"`). This is because we cannot extract dates from columns of deleted files. If hub submission window reference dates do not match round IDs in file paths, currently `allow_submit_window_mods` will not work correctly and is best set to `FALSE`. This only relates to hubs/rounds where submission windows are determined relative to a reference date and not when explicit submission window start and end dates are provided in the config.

Up to now and by default in ths PR, relative submission windows have been determined using contents of the file (which follows how the config files are set up to give the reference column name relative to which the window is determined). So far this date always matched the round ID in the file path so I opted for the option above to get round the problem described. 
This does mean that two submission checks (one when we check whether the submission falls within the submission window as part of `validate_submission` (which is also used separately, locally by users) and once now when `allow_submit_window_mods` = `TRUE` and mods/deletions are detected) use two different sources of reference date to calculate the window (the former uses the file contents to follow how config is structured more faithfully and allow more flexibility while the later the file path to allow for checking deleted file windows). 

I obviously don't love this duplication and difference in ref date source but I dislike the thought of introducing irrelevant arguments to `validate_submission` in order to pass upstream arguments from `validate_pr` as well introducing functionality to handle deleted files in `validate_submission` more. 

I feel the separation of the task of checking and alerting about mods & deletions to `validate_pr()` is correct and I've tried to handle the idiosyncrasies of checking for submission windows on deleted files as best I could, especially through documentation but if you have a better suggestion or can envisage a big problem I've overlooked let me know!

### DIFF
--- a/R/validate_pr.R
+++ b/R/validate_pr.R
@@ -271,5 +271,6 @@ check_pr_modf_del_files <- function(pr_df, file_type = c(
       )
     )
   }
-  as_hub_validations(out)
+  as_hub_validations(out) %>%
+    purrr::set_names(sprintf("%s_mod_%i", file_type, seq_along(out)))
 }
--- a/R/validate_pr.R
+++ b/R/validate_pr.R
@@ -8,7 +8,7 @@
 #' `round_id`s. Only required if files contain a column that contains `round_id`
 #' details but has not been configured via `round_id_from_variable: true` and
 #' `round_id:` in in hub `tasks.json` config file.
-#' @param file_modify_check Character string. Whether to perform check and what to
+#' @param file_modification_check Character string. Whether to perform check and what to
 #' return when modification/deletion of a previously submitted model output file
 #' or deletion of a previously submitted model metadata file is detected in PR:
 #' - `"error"`: Appends a `<error/check_error>` condition class object for each
@@ -28,11 +28,12 @@
 #' files are also validated. Any other files included in the PR are ignored but
 #' flagged in a message.
 #'
-#' By default, modifications and deletions of previously submitted model output
-#' files and deletions of previously submitted model metadata files are not allowed
+#' By default, modifications (which include renaming) and deletions of
+#' previously submitted model output files and deletions or renaming of
+#' previously submitted model metadata files are not allowed
 #' and return a `<error/check_error>` condition class object for each
 #' applicable modified/deleted file. This behaviour can be modified through
-#' arguments `file_modify_check`, which controls whether modification/deletion
+#' arguments `file_modification_check`, which controls whether modification/deletion
 #' checks are performed and what is returned if modifications/deletions are detected,
 #' and `allow_submit_window_mods`, which controls whether modifications/deletions
 #' of model output files are allowed within their submission windows.
@@ -61,13 +62,13 @@
 validate_pr <- function(hub_path = ".", gh_repo, pr_number,
                         round_id_col = NULL, validations_cfg_path = NULL,
                         skip_submit_window_check = FALSE,
-                        file_modify_check = c("error", "warn", "message", "none"),
+                        file_modification_check = c("error", "warn", "message", "none"),
                         allow_submit_window_mods = TRUE,
                         submit_window_ref_date_from = c(
                           "file",
                           "file_path"
                         )) {
-  file_modify_check <- rlang::arg_match(file_modify_check)
+  file_modification_check <- rlang::arg_match(file_modification_check)
   model_output_dir <- get_hub_model_output_dir(hub_path)
   model_metadata_dir <- "model-metadata"
   validations <- new_hub_validations()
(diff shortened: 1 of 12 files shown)


gioco:
Hi, we are experiencing an issue related to this. 
Contributing teams shall be allowed to modify their submission while within the submission window. Is there a way to allow it with the former version of the code (the one in the main branch)? 
If not, can you provide an estimate about when the PR will be merged and the allow_submit_window_mods parameter enabled? 
Thank you very much!

annakrystalli:
Hi @gioco ! Hopefully this will be merged within the next couple of days.

Out of interest, do you know what the issue your team's are having is? Are they submitting files, they are getting merged and they are committing modified versions within the submission window?

gioco:
Hi Anna, yes exactly. 
We have a workflow that automatically merges the submission if all validations are passed. 
A team submitted forecasts yesterday, and today wants to update them. Submission window closes tomorrow, and we want updates to be allowed while the submission window is still "open".
Thank you very much for your prompt reply.

annakrystalli:
So yes, sadly, checks will fail currently so might need to do some manual merging just for a couple of days.

You could also see whether you could briefly set your Github Action to install the `hubValidations` package from this branch instead of the `main` branch and change it back when the PR is merged. A little risky though as this PR hasn't been reviewed yet.

gioco:
Will go with the manual merges for this couple of days! 
Thanks!

LucieContamin:
I like this new functionality, we often also have updates during a "round" of submission. It seems to be all good, the function documentation is clear. Thanks @annakrystalli !

R/validate_pr.R
```diff
@@ -1,14 +1,52 @@
...
+#' `validate_submission()` with each file although as part of checks, hub config
+#' files are also validated. Any other files included in the PR are ignored but
+#' flagged in a message.
+#'
+#' By default, modifications and deletions of previously submitted model output
+#' files and deletions of previously submitted model metadata files are not allowed
+#' and return a `<error/check_error>` condition class object for each
+#' applicable modified/deleted file. This behaviour can be modified through
+#' arguments `file_modify_check`, which controls whether modification/deletion
+#' checks are performed and what is retuned if modifications/deletions are detected,
```
elray1, line 36:
```suggestion
#' checks are performed and what is returned if modifications/deletions are detected,
```

elray1 (commented):
Thanks! Here is a summary:

- one one-character typo :)
- how is all of this controlled?  Should we add documentation describing relevant settings (if any) that can be specified in a hub's validations config? OK with me if we add a separate issue for this so we can merge this in sooner.
- I have a request to add a little more documentation in the future -- it took some time for me to work through what was happening in `check_pr_modf_del_files`, and this would have been easier with just a few comments. (it's ok with me if you don't go back and do this now)

elray1 (approved):

annakrystalli:
- Thanks for the typo correction!
- Very fair point about docs. I was actually planning to spend all of tomorrow adding more docs to functions as well as vignettes to address #64 as well as how to configure validation runs in GitHub Actions using the arguments in `validate_pr`. I believe these docs belong in vignettes in this package but can certainly link to the appropriate hub config sections in hubDocs. If you have any specific suggestions on what you'd like to see in such docs, feel free to add them to #64
- No problem! Will aim to add more comments to internal functions.

annakrystalli:
Hello again @elray1 !

As we're planning to push this through quickly, I spent some time thinking ahead to exceptions and refactored `validate_pr` and internal functions to be more robust, specifically by handling errors in calculating submission windows at the individual file level, so that it doesn't cause an execution error at `validate_pr` level. I also added more better/tests.

I made some other minor improvements and added more comments so hopefully, between that and the refactoring, it should be much easier to follow what's going on.

If you could take a quick look at just this file: https://github.com/Infectious-Disease-Modeling-Hubs/hubValidations/blob/474f8b66c661696945310535dbb542ec6fa7437c/R/validate_pr.R 

I'd really appreciate it. Then we can go ahead and merge in! 🎉

elray1 (approved):
I read through again, looks good.  thanks!

PR #87, opened 2024-06-06
84/ignore compress ext

### LINKED ISSUE #84: `validate_submission()` does not accept filename with compression (for example:  `.gzip.parquet`
It seems that when `arrow` is writing a compressed parquet file the compression can or cannot be included in the filename, both will work. 
For example: 
`arrow::write_parquet(df, "model-output/JHU_UNC-flepiMoP/2024-04-28-JHU_UNC-flepiMoP.gzip.parquet", compression = "gzip", compression_level = 9)` and `arrow::write_parquet(df, "model-output/JHU_UNC-flepiMoP/2024-04-28-JHU_UNC-flepiMoP.parquet", compression = "gzip", compression_level = 9)`, both returns the same file with the same content and is possible to read the files with the same arrow function call. 

However, when using the `validate_submission()` , if we use a filename with the compression information, it will returns an error and not validate the files:
```
✖ 2024-04-28-JHU_UNC-flepiMoP.gz.parquet: EXEC ERROR: Error in parse_file_name(file_path) : Could
  not parse file name 2024-04-28-JHU_UNC-flepiMoP.gz for submission metadata. Please consult
  documentation for file name requirements for correct metadata parsing.
```

annakrystalli:
File names are allowed to contain the following compression extension prefixes: .snappy, .gzip, .gz, .brotli, .zstd, .lz4, .lzo, .bz2. These extension prefixes are now extracted by `parse_file_name()` when parsing the file name and returned as `compression_ext` element if present.

Resolves #84


LucieContamin (approved):
Looks good to me!

Issue #88, opened 2024-06-17
Validate coarse-grained samples


annakrystalli:
Currently, samples must strictly match the compound task ID set expectations and cannot handle coarser-grained compound task ID sets. However, [in our documentation we state that this should be allowed.](https://hubverse.io/en/latest/user-guide/sample-output-type.html#configuration-of-output-type-id)

This will need more work to implement but it would be good to understand the purpose of this and details of implamentation.

### Questions:

1. Can different teams be submitting completely different coarser grained samples (i.e. completely different compound idxs?)
2. Must the compound idx be consistent at least through an individual sample?
3. Does this affect the way we infer sample dependence (e.g. https://github.com/Infectious-Disease-Modeling-Hubs/hubData/issues/6)? i.e. will different samples be allowed to have different dependence structure? Will it affect plotting in any way? https://github.com/Infectious-Disease-Modeling-Hubs/hubVis/issues/31

annakrystalli:
@elray1 , @nickreich and @LucieContamin would appreciate your thoughts!

LucieContamin:
Please find here my answers to your questions:

1. My understanding is that in a same round, teams can submit files with different compound ids but they should at least include the minimal level expected from the hub. For example, if a hub has a round defined with 6 task id columns: `"origin_date"`, `"scenario_id"`, `"location"`, `"target"`, `"age_group"` and `"horizon"` with `"compound_taskid_set": ["origin_date", "scenario_id", "location", "target"]`. A team can submit a file with `"compound_taskid_set": ["origin_date"]` and another with `"compound_taskid_set": ["origin_date", "scenario_id", "target"]`. However, a file with `"compound_taskid_set": ["origin_date", "age_group"]` should not be accepted.

2. I am not sure I totally understand the question here. For me, an individual sample should have a consistent compound id that match the minimal level expected from the hub. And all the samples should follow the same compound id level. 

3. In my experience, to have different model output file for the same round with different level of "sample dependence" does impact the visualization but it's easy to adapt to it but I guess it depends what you want to do with the model output (same for data processing). Once you infer the structure of the samples of the files, it's easy to adapt as necessary, especially as all the model output should at least follow the same minimal structure. 

4. I don't understand that one, so will wait the answers from the others!

annakrystalli:
I would like to start work on this and have come up with a mechanism to implement it.
However, there are still things I am confused about.

Specifically, I can see problems with:
[...]

PR #95, opened 2024-07-08
Complete v3 sample validations / prepare optimisation work

annakrystalli:
In this package version I will complete the missing validation steps (e.g #88 ) as well as do some related preparation to optimise validations.

The first step is to move and rename the following `hubData` functions to `hubValidations`:
 - `hubData::expand_model_out_val_grid` to `expand_model_out_grid`.
 - `hubData::create_model_out_submit_tmpl` to `submission_tmpl`.

`hubValidations` is a much more natural place for these functions as `expand_model_out_val_grid()` is exclusively used internally by `hubValidation` check functions while `create_model_out_submit_tmpl` seems to also fit better as a helper for creating template files that can then be completed by teams and then validated locally before submission. Hence `hubValidations` feels like the appropriate home for this function too. I'm also taking the opportunity to rename the two functions to something more concise.

This PR simply moves and renames these two functions as well as any associated test materials to `hubValidations` and is related to the associated PR in `hubData`  https://github.com/hubverse-org/hubData/pull/46


nickreich (approved):
I didn't re-run the code to validate it after porting over from hubData, but the changes look good to me.

Issue #85, opened 2024-06-03
Memoise `expand_model_out_vals_grid`


annakrystalli:
As config files become more complex, this function becomes more computationally expensive.

Because this function effectively evaluates to the same output everytime for a given hub (unless of course the config changes), it's a good candidate for [memoisation](https://memoise.r-lib.org/), especially for use in a function to validate many files in a hub (#83 )

If implemented, a note in the docs should also be included to alert users that if they make changes to the config in between calls they should refresh their session to clear the cache.

PR #99, opened 2024-07-11
[WIP] Feature/handle coarse spls

annakrystalli:
This is the first part of work related to allowing for validation of coarser samples than those specified in a modeling tasks `compound_taskid_set` (#88), discussed in [this section of hubDocs](https://hubverse.io/en/latest/user-guide/sample-output-type.html#configuration-of-output-type-id).

The PR adds a new check function, `check_tbl_spl_compound_taskid_set()` to `validate_model_data()` which checks that sample compound task id sets for each modeling task is a single unique set that matches or is coarser than the expected set defined in the config. If the check fails, it will cause an early return of `validate_model_data()` as further validation of sample data is not possible.

To perform the above check, in this PR I'm also adding some utilities for detecting the compound task ID structure of samples in a submission file.  This functionality will then be added to the rest of the sample checks so that they all use the `compound_taskid_set` detected from the samples rather than the fixed set defined in the config.  

There were a couple of complications in implementing the `compound_taskid_set` detection discussed in #88 ( in particular https://github.com/hubverse-org/hubValidations/issues/88#issuecomment-2214295977 & https://github.com/hubverse-org/hubValidations/issues/88#issuecomment-2214455575) and I've implemented the solution described in https://github.com/hubverse-org/hubValidations/issues/88#issuecomment-2222467279

The PR also the message in `skip_v3_spl_check()`  so that the function that actually called the function and is being skipped is recorded in the output.


R/compound_taskid-utils.R
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  if (nrow(tbl) == 0L) {
+    cli::cli_alert_info("No samples found in the file.")
+    return(NULL)
+  }
+  if (is.null(call)) call <- rlang::current_env()
+  out_tid <- hubUtils::std_colnames["output_type_id"]
+  task_ids <- hubUtils::get_task_id_names(config_tasks)
+
```
annakrystalli, line 75: For extra context on the following code see https://github.com/hubverse-org/hubValidations/issues/88#issuecomment-2222467279

R/compound_taskid-utils.R
```diff
@@ -0,0 +1,203 @@
...
+#' @param compact Logical. If TRUE, the output will be compacted to remove NULL elements.
+#' @param error Logical. If TRUE, an error will be thrown if the compound task ID set is not valid.
+#' If FALSE and an error is detected, the detected compound task ID set will be
+#' returned with error attributes attached.
+#'
+#' @return A list of vectors of compound task IDs detected in the tbl, one for each
+#' modeling task in the round. If `compact` is TRUE, modeling tasks returning NULL
+#' elements will be removed.
+#' @noRd
+get_tbl_compound_taskid_set <- function(tbl, config_tasks, round_id,
```
annakrystalli, line 16: Do you think it would be useful to export this function?

LucieContamin (commented):
It looks good, thank you! I really like the way you extract the compound id set from the submission. 

However, I have a question. If I test this version of the validation with a submission file with a coarser compound id set, I still get an error. Is that expected? As I understand, it's still a work in progress but wanted to be sure it's expected.

The error is related to the function `check_tbl_spl_compound_tid ()`, for example:
```
2024-04-28-JHU_UNC-flepiMoP_subset.parquet: Each sample compound task ID does not contain single,
  unique value.  Samples "1", "100", "103", "106", "112", "119", "12", "120", "123", "125", "133",
  "135", "138", "141", "142", "145", "148", "152", …, "90", and "94" do not contain unique compound
  task ID combinations. See `errors` attribute for details.
```

annakrystalli:
@LucieContamin thanks for taking a look and testing it out! And yes, the validation of coarser samples have not propagated to all tests yet. This is step 1, then #100 is step 2 and once we are happy with both PRs I will go ahead and combine the functionality in the two PRs to allow me to propagate the create coarser sample compound idxs for the rest of the the tests.

annakrystalli:
So I will not merge this into `main` untill all functionality has been merged into this branch.

LucieContamin (approved):
All good to me!

R/compound_taskid-utils.R
```diff
@@ -0,0 +1,202 @@
...
+      config_tasks = config_tasks,
+      round_id = round_id,
+      all_character = TRUE,
+      include_sample_ids = FALSE,
+      bind_model_tasks = FALSE
+    ),
+    function(.x) {
+      if (!has_spls_tbl(.x)) {
+        return(NULL)
+      }
```
elray1, line 37: This comment may fall under the category of premature optimization, but given recent discussion about how expand-gridding can be costly -- could we consider adding an argument to `expand_model_out_grid` to only do this for groups that have samples as an output type?

R/compound_taskid-utils.R
```diff
@@ -0,0 +1,202 @@
...
+      dplyr::inner_join(tbl, .x[, names(.x) != out_tid],
+        by = setdiff(names(tbl), out_tid)
+      )
```
elray1, line 40: related to the above comment but more generally, maybe we could avoid all expand-gridding by replacing the "expand grid and inner join" operation here with something where we loop through task groups within the round, and within each task group do a filter operation to keep rows where elements of all task id columns are in the combined optional and required values for that column within the group?

elray1: (i don't know whether this would be more or less efficient, just putting the thought out there)

R/compound_taskid-utils.R
```diff
@@ -0,0 +1,202 @@
...
+  if (!is.null(cols)) {
+    x <- x[, cols]
+  }
+  if (unique) {
+    x <- unique(x)
+  }
+  v <- apply(x, 1,
+    function(x) names(x)[x],
+    simplify = FALSE
+  )
```
elray1, line 185: I think this will generate an error if `length(cols) == 1`

R/compound_taskid-utils.R
```diff
@@ -0,0 +1,202 @@
...
+    purrr::map(~ dplyr::n_distinct(.x) > 1L) |>
+    tibble::as_tibble()
+  tbl_non_uniq[, config_comp_tids] <- TRUE
+
+  # Create a sample (output_type_id) vs compound_taskid membership boolean matrix.
+  # For a task id to be considered a compound_taskid, a unique value must be TRUE
+  # at the sample level and we need to ensure it's not a false positive.
+  # Here we combine the two checks.
+  is_compound_taskid <- apply(spl_uniq[, task_ids], 1,
+    function(x) x & tbl_non_uniq[, task_ids],
```
elray1, line 108: For this sample index, for each task id var, TRUE if both:

- there was only one value of that task id var for this sample; and
- the hub named it as a compound task id var OR (the hub didn't name it as a compound task id var AND the model submitted multiple values for it across different samples)

elray1: this makes sense.

elray1 (commented):
This looks good to me.  I had some questions about whether we could get away without expand_grid that you can feel free to ignore or file a separate issue if you want to think about them later.  I also had a comment about one function that may error in an edge case.

elray1:
Actually, just thought of another question related to the splitting of the submission data frame according to task groups.  If a model inappropriately uses the same sample index across tasks that correspond to values in two groups, will that be detected?

R/compound_taskid-utils.R
annakrystalli: There is no other straight forward way to ensure checking against the right modeling task group config.
However, this is a key area I'm going to be focusing during optimisation work, coming up next. The combination of #98 & #85 means that additional expand grid calls will not add to execution time.

R/compound_taskid-utils.R
annakrystalli: That's the plan! See https://github.com/hubverse-org/hubValidations/issues/98

annakrystalli:
> Actually, just thought of another question related to the splitting of the submission data frame according to task groups. If a model uses the same sample index across tasks that correspond to values in two groups, will that be detected? e.g. what if the hub set their config up with one task group for flu and one task group for covid, and lists pathogen in their compound task id set. but a modeler's samples are "joint across" those. will we see that?

Not in this particular check but it will surely break other sample related checks, e.g. `check_tbl_spl_compound_tid()`, `check_tbl_spl_n()` or `check_tbl_spl_non_compound_tid()`.

R/compound_taskid-utils.R
annakrystalli: Exported in 8f39733dfadd4723b2d8bba04d0ad34c6e2f41dc!

R/compound_taskid-utils.R
annakrystalli: Thanks! Well caught! I'm so used to working with `tibbles` I've forgotten to check for this behavour! Fixed in 85e42576a392725f9dd3063e3ada4fa713703881

elray1 (approved):

PR #100, opened 2024-07-12
Override config compound_taskid_set

annakrystalli:
This is the second step in allowing all v3 sample checks to handle coarser `compound_taskid_set`s.

The PR introduces a `compound_taskid_set` argument to `expand_model_out_grid()` and `submission_tmpl()` that allows users to override the `compound_taskid_set` when creating sample indices in the `output_type_id` column of samples.

Combining the functionality in #99 for detecting `compound_taskid_set` from samples with the ability to use that to define coarser expected `compound_idx`s will allow the rest of the sample validation checks to accept and check against coarser sample structures (this last step to be introduced in a 3rd and last PR).

The introduction of the `compound_taskid_set` argument  to `submission_tmpl()` also allows users to see upfront what different expected  samples will look like for different `compound_taskid_set`s .


R/expand_model_out_grid.R
```diff
@@ -91,14 +94,27 @@
...
+#' # samples
+#' expand_model_out_grid(config_tasks,
+#'   round_id = "2022-12-26",
+#'   include_sample_ids = TRUE,
+#'   compound_taskid_set = list(
+#'     c("forecast_date", "target"),
+#'     NULL
+#'   )
+#' )
 expand_model_out_grid <- function(config_tasks,
```
LucieContamin, line 108: By trying the new parameter, I got an error message if I used:
```r
compound_taskid_set = list(
                            c("origin_date"),
                            NULL
                      )
```

I fixed it by changing line 417 of this file:
from ` spl <- unique(spl[, comp_tids]) %>%` to
  `spl <- unique(spl[, comp_tids, FALSE]) %>%`, to be sure to keep the data frame output format

LucieContamin (commented):
It looks ok to me, I just have two changes, especially one to fix an error for compound task id set with one column.

R/expand_model_out_grid.R
```diff
@@ -303,7 +319,17 @@ null_taskids_to_na <- function(model_task) {
...
 # across multiple modeling task groups. Only apply to v3 and above sample output
 # type configurations.
-add_sample_idx <- function(x, round_config, config_tid) {
+add_sample_idx <- function(x, round_config, config_tid, compound_taskid_set = NULL) {
+  if (!is.null(compound_taskid_set) && length(compound_taskid_set) != length(x)) {
+    cli::cli_abort(
+      c("x" = "The length of {.var compound_taskid_set}
+      ({.val {length(compound_taskid_set)}})
+      must match the number of modeling tasks ({.val {length(x)}})
+        in the round."),
```
elray1, line 328: Similar question to the one I asked on the other PR: suppose a hub defines a config file with one task group for flu and another for covid, with values of the task id variable `pathogen` being different between the groups. one could imagine wanting to override the compound taskid set to not include `pathogen`, right?

R/expand_model_out_grid.R
elray1: will it work if I use an empty set for this list?

elray1 (commented):
It's the end of the day so this was a brief review -- but I added a couple of comments/questions.

R/expand_model_out_grid.R
LucieContamin: I will work if you set it to `NULL`

annakrystalli: It does work if `compound_taskid_set` in `NULL`. This is the default signaling to use the compound task Id set in the config.

It also works if one of the list elements in `NULL`. However the question made me think of what we may want out of that scenario. Whne the question was asked, providing a list with `NULL` for a given modeling task fell back to using the config compound task id set. However, after some thought, it felt like `NULL` in this situation should replicate the behaviour when `compound_taksid_set` is set to `NULL` for a given modeling task, i.e. use all task ids in `compound_taksid_set`. So I've gone ahead and implemented the above fix in ec42ae72d187196548acfee12a869112164d2da3.

R/expand_model_out_grid.R
annakrystalli: I'm not 100% sure I follow the question but as mentioned before, submitted data that refer to different modeling tasks are processed separately and can have their own `compound_taskid_set`. That's also why when we want to override them we provide a list with a character vector of compound task IDs for each modeling task.

annakrystalli:
@elray1 

Re:
> will it work if I use an empty set for this list?

No.

If you run:
```r
   expand_model_out_grid(
      config_tasks,
      round_id = "2022-12-26",
      include_sample_ids = TRUE,
      compound_taskid_set = list()
    )
```
 you get the following error:
```r
Error in `expand_model_out_grid()`:
✖ The length of `compound_taskid_set` (0) must match the number of modeling tasks (2) in the round.
```

Added another test in 52cda6d959f7fafe5e13b5f273d2198b5b8de503

NEWS.md
```diff
@@ -1,6 +1,7 @@
 # hubValidations 0.3.0
 
 - Add new `check_tbl_spl_compound_taskid_set()` check function to `validate_model_data()` that ensures that sample compound task id sets for each modeling task match or are coarser than the expected set defined in `tasks.json` config.
+- Add argument `compound_taskid_set` to `expand_model_out_grid()` and `submission_tmpl()` that allows users to override the `compound_taskid_set()` when creating sample indices in the `output_type_id` column of samples.
```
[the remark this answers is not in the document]
annakrystalli: Sorry, yeah the `()` `compound_taskid_set()`  was a typo but I added your suggestion to clarify.

elray1 (approved):

LucieContamin (approved):

PR #101, opened 2024-07-24
91/override out type id type

annakrystalli:
This PR Introduces an `output_type_id_datatype` argument to `validate_pr()`, `validate_submission()`, `validate_model_data()` and `expand_model_out_grid()` and set default value to `"from_config"`. 

This default means the data type specified in the `output_type_id_datatype` property in `tasks.json` (introduced in schema [version `v3.0.1`](https://github.com/hubverse-org/schemas/releases/tag/v3.0.1)) is used to cast the hub level `output_type_id` column data type. If not set in the config, the functions fall back to `"auto"` which detects the simplest data type that can represent all output type id values across all output types and rounds. The argument also allows hub administrators to override this setting manually during validation.


tests/testthat/test-check_tbl_col_types.R
```diff
@@ -20,15 +20,19 @@ test_that("check_tbl_col_types works", {
...
-  tbl <- read_model_out_file(
-    file_path,
-    hub_path
-  )
-  expect_snapshot(
-    check_tbl_col_types(tbl, file_path, hub_path)
-  )
 })
+
+test_that(
```
LucieContamin, line 25: As I understand, the behavior for CSV output as changed, should we add it in the documentation/NEWS?

R/read_model_out_file.R
```diff
@@ -1,16 +1,25 @@
 #' Read a model output file
 #'
 #' @inheritParams check_valid_round_id
+#' @inheritParams hubData::create_hub_schema
 #' @param coerce_types character. What to coerce column types to on read.
 #' - `hub`: read in (`csv`) or coerce (`parquet`, `arrow`) to hub schema.
```
LucieContamin, line 6: Should we had that it's the default selection?

LucieContamin (approved):
Looks all good to me! thanks. I just made some minor remarks for the documentation

tests/testthat/test-check_tbl_col_types.R
annakrystalli: Sorry I'm not sure I follow. We've added the options to configure the value of this argument through the config but if that's not set (so all hubs up to now) it falls back to the previously default behaviour of auto-detection. So nothing really has changed unless you upgrade your config version and use the `output_type_id_datatype` property. Also not sure what you mean re: CSVs specifically. There hasn't been any format specific changes.

R/read_model_out_file.R
annakrystalli: Sure. Fixed in 3e9b0f4c0b526e34a59516a4e7222d1fb9059bd3

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# hubValidations 0.3.0
+
+* Introduce an `output_type_id_datatype` argument to `validate_pr()`, `validate_submission()`, `validate_model_data()` and `expand_model_out_grid()` and set default value to `"from_config"`. This default means the data type specified in the `output_type_id_datatype` property in `tasks.json` (introduced in schema version `v3.0.1`) is used to cast the hub level `output_type_id` column data type. If not set in the config, the functions fall back to `"auto"` which detects the simplest data type that can represent all output type id values across all output types and rounds. The argument also allows hub administrators to override this setting manually during validation.
```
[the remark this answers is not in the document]
annakrystalli: I think it's ok honestly because there's a lot more detail in the actual function documentation and I don't think adding _automatically_ there adds that much more clarity. Feels more like repetition.

annakrystalli:

PR #103, opened 2024-08-06
Propagate detected compound_taskid_set throughout rest of spl checks

annakrystalli:
This PR adds the final part of work related to validating samples with coarser compound task id sets.

Building on the functionality introduced in https://github.com/hubverse-org/hubValidations/pull/99 for detecting compound_taskid_set from samples and the introduction of a `compound_taskid_set` argument to `expand_model_out_grid()` in #100 that allows users to override the `compound_taskid_set` when creating sample indices in the `output_type_id` column of samples, allows the rest of the sample validation checks to accept and check against coarser sample structures.

This is the functionality introduced in this 3rd and last PR.

More specifically, the `compound_taskid_set` detected and validated in the first `check_tbl_spl_compound_taskid_set()` check is included as part of the check result (if valid). If `check_tbl_spl_compound_taskid_set()` passes, the `compound_taskid_set` is then passed on to downstream sample checks and overrides the `compound_taskid_set` defined in the config. 

This way:
1. there is a record of the `compound_taskid_set` used for validation in the output of the `check_tbl_spl_compound_taskid_set()` 
2. the `compound_taskid_set` is determined once per validation run and passed on to other checks.

Another addition to this PR is a missing check for consistent number of samples across compound IDs, which is now included as part of `check_tbl_spl_n()`


R/check_tbl_spl_n.R
```diff
@@ -14,21 +15,56 @@
...
+  hash_tbl <- spl_hash_tbl(tbl, round_id, config_tasks, compound_taskid_set)
   n_ranges <- get_round_spl_n_ranges(config_tasks, round_id)
 
   n_tbl <- dplyr::group_by(hash_tbl, .data$compound_idx) %>%
     dplyr::summarise(
       n = dplyr::n_distinct(.data$output_type_id),
       mt_id = unique(.data$mt_id)
-    ) %>%
+    )
+
```
annakrystalli, line 43: The following splits the previous check that only looked at whether samples for each compound idx fall within the expected range into:
1. check that all  compound idxs have the same number of samples
2. If above check passes, check that the number of samples falls within the expected range

R/v3-sample-utils.R
```diff
@@ -6,58 +6,75 @@
...
+          dplyr::rename(compound_idx = "output_type_id") %>%
+          dplyr::inner_join(tbl, .,
+            by = setdiff(
+              names(tbl),
+              c("output_type_id", "compound_idx")
+            )
+          )
+      }
+    }
+  )
```
annakrystalli, line 41: The above is a refactoring of the below that allows us to not have to subset different modeling tasks but instead do nothing when there are no samples in a given modeling task. Overall this makes downstream step cleaner.

R/validate_model_data.R
```diff
@@ -196,13 +196,16 @@ validate_model_data <- function(hub_path, file_path, round_id_col = NULL,
     )
     if (is_any_error(checks$spl_compound_taskid_set)) {
       return(checks)
+    } else {
+      compound_taskid_set <- checks$spl_compound_taskid_set$compound_taskid_set
```
annakrystalli, line 200: We first validate and detect the compound task id set and if it validates, we pass the compound task id set to subsequent tests.

LucieContamin (approved):
I test it quickly and read the codes and it looks all good to me, thank you very much! I just let a small comment on a comment that can be remove now I think.

R/check_tbl_spl_compound_tid.R
```diff
@@ -13,14 +16,24 @@
...
+    )
+  }
 
   if (isFALSE(has_spls_tbl(tbl)) || isFALSE(hubUtils::is_v3_config(config_tasks))) {
     return(skip_v3_spl_check(file_path))
   }
 
-  hash_tbl <- spl_hash_tbl(tbl, round_id, config_tasks)
+  hash_tbl <- spl_hash_tbl(tbl, round_id, config_tasks, compound_taskid_set)
   # TODO: Currently, samples must strictly match the compound task ID set expectations
```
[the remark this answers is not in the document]
annakrystalli: Well spotted!. Removed

annakrystalli:

Issue #93, opened 2024-07-02
Improve validation performance


annakrystalli:
## Background

While our current validation functionality works well enough on smaller files / hubs with less complex config files, it can be much slower on larger files / more complex config. This has also been noted and reported by the community (e.g. #86)

The most time consuming functions are checking that the combination of values are valid and validating that all required value combinations have been submitted. 

There are  a number of reasons/bottlenecks 
- The size of the expanded grid of all possible values in a complex hub can be very large.
- The expansion of value combinations which are effectively invalid because their value is related to the value of another variable (e.g `target_end_date` which only has a single valid value dependant on `origin_date` and `horizon` - see hubverse-org/hubValidations#38.) Expanding the values of such task ids unnecessarily increases the size of the expanded value grid while the actual validation is performed via optional validation check `hubValidations::opt_check_tbl_horizon_timediff()`.
- Creating an index via [`conc_rows`](https://github.com/hubverse-org/hubValidations/blob/bfab4d31e43e1e3cce2f4854e0c02f9694038a9c/R/conc_rows.R#L1-L14) to then [split the submitted table](https://github.com/hubverse-org/hubValidations/blob/bfab4d31e43e1e3cce2f4854e0c02f9694038a9c/R/check_tbl_values_required.R#L85-L86) and check for required values in `check_tbl_values_required`.

These are likely the most effective areas to direct effort to improve performance.

## Specific Actions

- [x] **Perform memory intensive validations in a piecemeal way:** Once https://github.com/hubverse-org/hubValidations/issues/98 is implemented, we should refactor any checks making use of `expand_model_out_val_grid()` to perform the checks one output type at a time. This way we avoid burdening memory with the full expanded grid at any one time.
- [x] **Memoise `expand_model_out_val_grid()`**: As this function is called a number of times but always returns the same result for the same config, it's a good candidate for memoisation (#85) .
- [x] **Optimise `conc_rows`:** I've already tried and failed to improve the performance of this function but given it's the main bottleneck to `check_tbl_values_required`, it feels important to revisit and try again.
- [x] **Introduce mechanism for excluding task ids from expanded grid of valid values**: This relates to task ids like `target_end_date` in which expanding their values is meaningless yet can be very memory consuming. For such task ids, validation would involve:
    - validating that the unique values in the task id column are valid with respect to the config (instead of checking them as part of combinations)
    - Using custom/optional functions to validate expected properties/relationships of such variables.
[...]

PR #108, opened 2024-08-14
93/add derived_task_id argument

annakrystalli:
This is the second PR associated with improving the performance of validation workflows ( #93 ) and provides a mechanism for ignoring derived task ids which can not only pollute expanded grids of valid values with effectively non valid value combinations but also significantly increase the size of expanded grids unnecessarily and affect validation performance.

To allow us to by pass such task ids this, I'm introducing a `derived_taks_ids` argument in `expand_model_out_grid()` which ignores any task id names provided when expanding the valid value grid, returning such columns as `NA`.

I've also moved some utility functions to a separate R script (`R/config_tasks-utils.R`) but nothing in these fns has actually changed

### DIFF
--- a/R/config_tasks-utils.R
+++ b/R/config_tasks-utils.R
@@ -1,0 +1,29 @@
+get_round_config <- function(config_tasks, round_id) {
+  round_idx <- hubUtils::get_round_idx(config_tasks, round_id)
+  purrr::pluck(
+    config_tasks,
+    "rounds",
+    round_idx
+  )
+}
+
+get_round_output_types <- function(config_tasks, round_id) {
+  round_config <- get_round_config(config_tasks, round_id)
+  purrr::map(
+    round_config[["model_tasks"]],
+    ~ .x[["output_type"]]
+  )
+}
+
+get_round_output_type_names <- function(config_tasks, round_id,
+                                        collapse = TRUE) {
+  out <- get_round_output_types(config_tasks, round_id) %>%
+    purrr::map(names)
+
+  if (collapse) {
+    purrr::flatten_chr(out) %>%
+      unique()
+  } else {
+    out
+  }
+}
--- a/R/expand_model_out_grid.R
+++ b/R/expand_model_out_grid.R
@@ -478,37 +478,6 @@ extract_mt_output_type_ids <- function(x, config_tid) {
   )
 }
 
-
-get_round_config <- function(config_tasks, round_id) {
-  round_idx <- hubUtils::get_round_idx(config_tasks, round_id)
-  purrr::pluck(
-    config_tasks,
-    "rounds",
-    round_idx
-  )
-}
-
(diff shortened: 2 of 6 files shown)


R/config_tasks-utils.R
```diff
@@ -0,0 +1,29 @@
...
+  purrr::map(
+    round_config[["model_tasks"]],
+    ~ .x[["output_type"]]
+  )
+}
+
+get_round_output_type_names <- function(config_tasks, round_id,
+                                        collapse = TRUE) {
+  out <- get_round_output_types(config_tasks, round_id) %>%
+    purrr::map(names)
```
elray1, line 21: I don't understand what this does just from reading the code, but I trust that it's reasonable.

R/expand_model_out_grid.R
```diff
@@ -345,6 +378,22 @@ null_taskids_to_na <- function(model_task) {
   )
 }
 
+# Set derived task_ids to all NULL values.
+derived_taskids_to_na <- function(model_task, derived_task_ids) {
+  if (!is.null(derived_task_ids) || length(derived_task_ids) > 0L) {
```
elray1, line 383: Did you want an `&&` in the condition here?  Again, I am just not sure what the inputs here might look like to be sure.  I could imagine `derived_task_ids = character(0)` passing this condition but being a surprise?

R/config_tasks-utils.R
elray1: I see we are just moving these functions to a separate file.

R/expand_model_out_grid.R
```diff
@@ -529,3 +547,47 @@ validate_output_types <- function(output_types, config_tasks, round_id,
   }
   valid_output_types
 }
+
+validate_derived_task_ids <- function(derived_task_ids, config_tasks, round_id) {
+  checkmate::assert_character(derived_task_ids, null.ok = TRUE)
+  if (is.null(derived_task_ids)) {
+  }
```
elray1, line 554: Empty code block; remove or was there content to be added here?

elray1 (approved):
only a few very minor comments.  will approve for merge as I don't think a re-review is needed, though it would be nice to address the question about the empty code block.

R/expand_model_out_grid.R
annakrystalli: IHmmm you're right, a `&&` statement would be more appropriate there but in all honesty, given the validation of `derived_task_ids` returns either a vector of valid task ID names or `NULL`, the second check is not necessary. Removed in c6c54237d22cdffc18d5eca811fd7331cd6c698b

R/expand_model_out_grid.R
annakrystalli: DOH!! Fixed in c6c54237d22cdffc18d5eca811fd7331cd6c698b 🙈

PR #117, opened 2024-09-06
Ak/clean up hub_validation print output. Deprecate file_modification_check argument "warn" 

annakrystalli:
This PR cleans up a few rough edges from the previous two PRs

* Changed file name header colour in `hub_validations` object `print()` method from `br_cyan` to `magenta` to make more visible on lighter backgrounds.
* Soft deprecated `file_modification_check` argument `"warn"` option and replaced it with `"failure"` in `validate_pr()` function.  Also I updated the output documentation so reflect the new `error/check_failure` class.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hubValidations
 Title: Testing framework for hubverse hub validations
-Version: 0.6.0
+Version: 0.6.1
 Authors@R: c(
     person(
         given = "Anna", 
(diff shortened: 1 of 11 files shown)


zkamvar (approved):
This looks good to me.

My only comment is to remove the coloring from the heading class since the surrounding dashes and underline distinguish it from the other content.

annakrystalli:
Thanks for the review @zkamvar ! I tried it with the colour removed but I feel it still adds some useful contrast so I've left it for now.

zkamvar:
> I tried it with the colour removed but I feel it still adds some useful contrast so I've left it for now.

It's minor, but it will be a good idea to revisit this decision because it may be an accessibility issue in the future.

Issue #121, opened 2024-09-24
Document requirements for custom validation functions


zkamvar:
# Summary

### Proposed outputs check list

- [x] `create_custom_check()` function (fixed in #122) 
- [x] Writing custom validation functions articles

## Initial background

I was testing out a potential custom validation function for https://github.com/reichlab/variant-nowcast-hub using the vignette in https://hubverse-org.github.io/hubValidations/articles/custom-functions.html#managing-dependencies-of-custom-sourced-functions

I am really glad to report that once I figured out the nuances of actually calling the new function, adding a custom function seems to be pretty plug-and-play! 

Here were the steps (with missing pieces labelled):

1. create a `validations.yml` file (NOTE: this can not be `validations.yaml` unless you want to manually modify your GitHub workflow)
2. (implied) create a script in `src/` that contains functions for validation, which contains the arguments specified in [the default configuration](https://hubverse-org.github.io/hubValidations/articles/custom-functions.html#default-configuration) (even if you don't use them)
3. **(missing) ensure that your function evaluates to `TRUE` or `FALSE` and then wrap the result in [`hubValidations::capture_check_cnd()`](https://hubverse-org.github.io/hubValidations/reference/capture_check_cnd.html)**
4. test locally by running `hubValidations::validate_submission()` **(missing: in the same directory of the hub)**

I tested this with the simple hub and the functions I created were:

```yaml
default:
    validate_model_data:
      hello:
        fn: "cromulent"
        source: "src/sum-to-one.R"
      sum1:
        fn: "sum_to_one"
        source: "src/sum-to-one.R"
```

```r
sum_to_one <- function(tbl = NULL, file_path, ...) {
  res <- isTRUE(all.equal(sum(tbl$value), 1, tol = 1e-5))
  hubValidations::capture_check_cnd(res,
    file_path = file_path,
    msg_subject = "b",
    msg_attribute = "c"
  )
}

cromulent <- function(tbl = NULL, file_path, ...) {
  hubValidations::capture_check_cnd(!is.null(tbl),
    file_path = file_path,
    msg_subject = "b",
    msg_attribute = "c"
  )
}
```

I would be happy to tackle this and update the vignette.

annakrystalli:
## 💯🚀 

Yes! This is definitely a gap in the docs I've not had time to tackle! Would be great if you wanted to work on this.

[...]

PR #122, opened 2024-09-27
Add `create_custom_check()` function

annakrystalli:
Add `create_custom_check()` function for creating custom validation check files from a template, related to #121 

Based largely on a combination of `usethis::use_r()` and `usethis::use_template()` functionality, the function creates an `<fn_name>.R` file in the default `src/validations/R/` directory (which it creates if it does not exist) and populates it with a useful function template designed to guide hub admins through easily creating their own custom validation functions.

Template contents can be parameterised via a number of function arguments and default file locations can also be overriden.

The templates use [mustache](https://mustache.github.io/mustache.5.html).


R/create_custom_check.R
```diff
@@ -0,0 +1,113 @@
...
+#' it could store the index of rows in a `tbl` that caused a check failure.
+#' @param extra_args Logical. If `TRUE`, the custom check function template will
+#' include an `extra_arg` template argument and template block of code to check
+#' the input arguments of the custom check function.
+#' @param overwrite Logical. If `TRUE`, the function will overwrite an existing
+#'
+#' @return Invisible `TRUE` if the custom check function file is created successfully.
+#' @details
+#' See the article on [writing custom check functions](
+#' https://hubverse-org.github.io/hubValidations/articles/writing-custom-fns.html)
```
annakrystalli, line 30: Note the article URL results in a 404 currently as it is still under development but will work in the next PR

inst/templates/custom_check_template
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+{{/conditional}}
+  # Here you can write your custom check logic
+  # Assign the result as `TRUE` or `FALSE` to object called `check`.
+  # If `check` is `TRUE`, the check will pass.
+
+  check <- condition_to_be_TRUE_for_check_to_pass
+
+if (check) {
```
zkamvar, line 27:
```suggestion
  if (check) {
```

inst/templates/custom_check_template
```diff
@@ -0,0 +1,51 @@
...
+  } else {
+{{#error_object}}
+    details <- cli::format_inline("See {.var error_object} attribute for details.")
+{{/error_object}}
+{{^error_object}}
+    details <- cli::format_inline("{.var round_id} value {.val invalid} is invalid.")
+{{/error_object}}
+  }
+
+  capture_check_cnd(
```
zkamvar, line 41:
```suggestion
  hubValidations::capture_check_cnd(
```

R/create_custom_check.R
```diff
@@ -0,0 +1,113 @@
...
+  checkmate::assert_character(name, len = 1L)
+  checkmate::assert_scalar(hub_path)
+  checkmate::assert_scalar(r_dir)
+  checkmate::assert_logical(error, len = 1L)
+  checkmate::assert_logical(conditional, len = 1L)
+  checkmate::assert_logical(error_object, len = 1L)
+  checkmate::assert_logical(extra_args, len = 1L)
+  checkmate::assert_logical(overwrite, len = 1L)
+
+  # If hub_path not current working directory, prefix path to R dir with hub_path.
```
zkamvar, line 61: Should add a check that the `hub_path` exists and is a hub.

R/create_custom_check.R
```diff
@@ -0,0 +1,113 @@
...
+# licensed under MIT License.
+# See https://github.com/r-lib/usethis/blob/2cc9e5a4e5e3034399fafb65d871eec352d83a15/R/template.R#L10
+create_template_fn <- function(data) {
+  template_path <- system.file("templates/custom_check_template",
+    package = "hubValidations"
+  )
+  readLines(template_path, encoding = "UTF-8", warn = FALSE) |>
+    whisker::whisker.render(data) |>
+    strsplit("\n") |>
+    unlist()
```
zkamvar, line 112: I don't believe it should be necessary to split by newline if this is just going to go directly to `writeLines()`. 

```suggestion
    whisker::whisker.render(data)
```

R/create_custom_check.R
```diff
@@ -0,0 +1,113 @@
...
+}
+
+# Much of the following functionality is heavily based on usethis function `use_template()`
+# licensed under MIT License.
+# See https://github.com/r-lib/usethis/blob/2cc9e5a4e5e3034399fafb65d871eec352d83a15/R/template.R#L10
+create_template_fn <- function(data) {
+  template_path <- system.file("templates/custom_check_template",
+    package = "hubValidations"
+  )
+  readLines(template_path, encoding = "UTF-8", warn = FALSE) |>
```
zkamvar, line 109: GOOD CATCH with the `warn = FALSE`! That annoying "no newline" warning has been the bane of my existence.

R/create_custom_check.R
```diff
@@ -0,0 +1,113 @@
...
+#' article on [writing custom check functions](
+#' https://hubverse-org.github.io/hubValidations/articles/writing-custom-fns.html).
+#' @param hub_path Character string. Path to the hub directory. Default is the
+#' current working directory.
+#' @param r_dir Character string. Path (relative to `hub_path`) to the directory
+#' the custom check function file will be written to. Default is `src/validations/R`
+#' which is the recommended directory for storing custom check functions.
+#' @param error Logical. If `TRUE`, the custom check function will return a
+#' `<error/check_error>` class object in the case of a failed check instead of the
+#' default `<error/check_failure>`.
```
zkamvar, line 14: I easily forget the differences between `check_failure` and `check_error` and I think it would be helpful to reiterate the differences between the two. 

```suggestion
#' @param error Logical. Defaults to `FALSE`, which will return a 
#' `<error/check_failure>` class object in the case of a failed check. 
#' Set this to `TRUE` if your custom check function is required to pass for 
#' other checks to be performed; the custom check function will then return a
#' `<error/check_error>` class object in the case of a failed check instead.
```

R/create_custom_check.R
```diff
@@ -0,0 +1,113 @@
...
+#' @param conditional Logical. If `TRUE`, the custom check function template will
+#' include a block of code to check a condition before running the check. This is useful
+#' when a check may need to be skipped based on a condition.
+#' @param error_object Logical. If `TRUE`, the custom check function template will
+#' include an error object that can be used to store additional information about the
+#' properties of the object being checked that caused check failure. For example,
+#' it could store the index of rows in a `tbl` that caused a check failure.
```
zkamvar, line 21: I was a little confused about this, but I think pointing to a worked example would really help.

inst/templates/custom_check_template
```diff
@@ -0,0 +1,51 @@
+{{name}} <- function(tbl, file_path{{#extra_args}}, extra_arg = NULL{{/extra_args}}) {
```
zkamvar, line 1: If I understand the custom function mechanism correctly, we will need to include all the arguments for `validate_model_data` (see https://hubverse-org.github.io/hubValidations/articles/custom-functions.html#default-configuration) so that we don't get `unused argument` errors

zkamvar (changes requested):
Thank you for tackling this, @annakrystalli! I still need to test this with a custom function, and have a couple of suggestions for documentation.

R/create_custom_check.R
annakrystalli: I've added a check to ensure a `hub_path` directory exists but don't want to be too prescriptive about the contents of `hub_path`. Not much is gained as a fully functional hub is not required for `create_custom_check()` to work correctly, it would make it difficult for folks to just create templates in a throwaway folder to play around with (including our own tests!) and there is no hard and fast rule for identifying a hub. We usually check for the part of the hub that is relevant to a given functionality which is not obvious here.

inst/templates/custom_check_template
annakrystalli: Not quite. All the objects listed in that vignette are available in the caller function and can be passed **if required by any custom function**.

annakrystalli:
Thanks for an initial pass @zkamvar ! I committed/ resolved or responded to your comments.

Your comments also prompted me to revisit the functionality for executing custom functions and ensure that custom checks indeed returned early if any custom checks returned a `<error/check_error>` or exec error class object. I added a bunch more tests too.

I also fixed a couple of small bugs I found and added an additional `config` argument for inserting a required argument and command for loading a tasks.json config file to the template.

On an aside, I noticed the new release docuemtation process has broken our PR documentation previews. I've opened and issue here: https://github.com/hubverse-org/hubverse-developer-actions/issues/4 and had a preliminary look for a fix but haven't quite got there. If you have any ideas, feel free to chime in.

inst/templates/custom_check_template
annakrystalli: I'm going to pull that section of the article into a child doc we can repeat in both and will try to clarify that point

tests/testthat/helper-custom-fns.R
```diff
@@ -0,0 +1,13 @@
+test_custom_checks_caller <- function(
+    # nolint start
```
zkamvar, line 2: I fully support this rebellion against the tyranny of the linter

tests/testthat/testdata/src/R/src_check_horizon_timediff.R
```diff
@@ -0,0 +1,86 @@
...
+
+assert_column_date <- function(colname, schema) {
+  if (schema[colname] != "Date") {
+    cli::cli_abort(
+      "Column {.arg colname} must be configured as {.cls Date} not
+      {.cls {schema[colname]}}.",
+      call = rlang::caller_call()
+    )
+  }
+}
```
zkamvar, line 86: **non-blocking suggestion**

The tests use this function to test for a check that uses `error = TRUE`. This function achieves that, but it does so by copying over the logic from an existing check that may change in the future.

Instead, I suggest we should make this function always return `FALSE`. This way, when the original function changes in the future, we don't have to update this one. 


```suggestion
    check <- FALSE 
  if (check) {
    details <- NULL
  } else {
    invalid_vals <- get_invalid_vals()
    details <- cli::format_inline(
      "t1 var value{?s} {.val {invalid_vals}} are invalid."
    )
  }

  hubValidations::capture_check_cnd(
    check = check,
    file_path = file_path,
    error = TRUE,
    msg_subject = cli::format_inline(
      "Time differences between t0 var {.var {t0_colname}} and t1 var
        {.var {t1_colname}}"
    ),
    msg_verbs = c("all match", "do not all match"),
    msg_attribute = cli::format_inline("expected period of {.val {timediff}} * {.var {horizon_colname}}."),
    details = details
  )
}

get_invalid_vals <- function() {
  c("2023-05-15 (horizon = 1)", "2023-05-22 (horizon = 2)")
}
```

inst/templates/custom_check_template
zkamvar: Thank you for the clarification! I will admit that I didn't look too deeply at the code evaluating the custom functions and I am impressed!

zkamvar (approved):
I think this is fantastic! I'm excited to see what you come up with in the vignette.

> Your comments also prompted me to revisit the functionality for executing custom functions and ensure that custom checks indeed returned early if any custom checks returned a <error/check_error> or exec error class object. I added a bunch more tests too.

YESSS MORE TESTS!

tests/testthat/testdata/src/R/src_check_horizon_timediff.R
annakrystalli: This is a great point! In fact I was being a bit lazy and taking the shortest path to a working testable example. I've actually refactored and simplified the materials associated with these tests to focus solely on the interaction of different conditions with custom check execution. Also added tests for early return as a result of execution errors

Issue #123, opened 2024-10-02
`check_tbl_values_required()` fails to detect that all required task IDs are present


zkamvar:
An issue was found in https://github.com/reichlab/variant-nowcast-hub/issues/83, which is tested in https://github.com/reichlab/variant-nowcast-hub/pull/90 where `check_tbl_values_required()` does not invalidate the submission when required task ID values are missing in a hub that is taking sample data. 

``` r
tmp <- tempfile()
tmpdir <- tempfile()
dir.create(tmpdir)
download.file("https://github.com/IsaacMacarthur/variant-nowcast-hub/archive/b80f0559c7fd019b569f39437cf684aca57639a6.zip", tmp)
unzip(tmp, exdir = tmpdir)

library("hubValidations")
library("dplyr")
#> 
#> Attaching package: 'dplyr'
#> The following object is masked from 'package:hubValidations':
#> 
#>     combine
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
hub_path <- list.files(tmpdir, full.names = TRUE)
file_path <- "UMass-HMLR/2024-10-02-UMass-HMLR.parquet"
round_id <- "2024-10-02"
tbl <- read_model_out_file(hub_path = hub_path, file_path = file_path) %>%
  mutate_at(vars(ends_with("date")), as.character)
```

There are supposed to be specific clades in the task IDs

``` r
tasks <- hubUtils::read_config_file(file.path(hub_path, "hub-config", "tasks.json"))
purrr::map(tasks$rounds, list("model_tasks", 1, "task_ids", "clade"))
#> [[1]]
#> [[1]]$required
#> [1] "24A"         "24B"         "24C"         "other"       "recombinant"
#> 
#> [[1]]$optional
#> NULL
#> 
#> 
#> [[2]]
#> [[2]]$required
#> [1] "24A"         "24B"         "24C"         "other"       "recombinant"
#> 
#> [[2]]$optional
#> NULL
#> 
#> 
#> [[3]]
#> [[3]]$required
#> [1] "24A"         "24B"         "24C"         "other"       "recombinant"
#> 
#> [[3]]$optional
#> NULL
#> 
#> 
#> [[4]]
#> [[4]]$required
#> [1] "24A"         "24B"         "24C"         "24E"         "other"      
#> [6] "recombinant"
#> 
#> [[4]]$optional
#> NULL
```

[...]

PR #124, opened 2024-10-03
Fix required vals check bug

annakrystalli:
The bug is described in #123 and solved partly by @zkamvar 's suggestion in the issue.

Fixed fully by differentiating behaviour of `expand_model_out_grid()`  when required_vals are requested and all output types are optional and when  required_vals are requested for subsets of output types.

Added detailed comment in code that explains the reasoning.

Also added a small nowcast example hub.

Note I've incremented the version number for tracking of this bug.


tests/testthat/test-check_tbl_values_required.R
```diff
@@ -251,3 +251,33 @@ test_that("Ignoring derived_task_ids in check_tbl_values_required works", {
     )
   )
 })
+
+test_that("check_tbl_values_required works with all optional output types", {
```
zkamvar, line 255: Adding the issue number for the test so that we can be alerted to the context immediately if something breaks again. 

```suggestion
test_that("(#123) check_tbl_values_required works with all optional output types", {
```

zkamvar (approved):
Thank you for the quick work, Anna!

I'm glad to have your eyes on this because I did not even think about the potential of output type subsetting (as evidenced by the failing tests after my fix)!

I am approving this, and as usual, I have two suggestions:

1. add the issue number to the new test so that we can track it quickly if it fails again
2. add another finer-grain unit test for the functionality change in `tests/testthat/test-expand_model_out_grid.R`:

```r
test_that("(#123) expand_output_type_grid() returns expected outputs with optional output_type_id", {
  tasks <- list(
    nowcast_date = "2024-10-02",
    target_date = NULL,
    clade = c("24A", "24B", "recombinant", "other"),
    other_task = 1:2,
    location = NULL
  )
  # If specific output type subset is requested
  i_have_no_rows <- expand_output_type_grid(
    task_id_values = tasks,
    output_type_values = list(),
    all_output_types = FALSE
  )
  expect_equal(nrow(i_have_no_rows), 0)
  expect_equal(ncol(i_have_no_rows), 0)

  # When no specific output_type subset is requested
  i_have_eight_rows <- expand_output_type_grid(
    task_id_values = tasks,
    output_type_values = list(),
    all_output_types = TRUE
  )
  expect_equal(nrow(i_have_eight_rows), 8)
  expect_equal(ncol(i_have_eight_rows), 3)

})
```


In terms of process, even though the vignette for #121 is not yet complete, **I think we should create a release after this PR** since this is needed for the variant nowcast hub to go live (though that's not _necessarily_ true... we could patch it in the custom functions).

annakrystalli:
Thanks @zkamvar .

I've committed, updated the snapshots and add the new tests. Will merge when tests pass

zkamvar:
> I've committed, updated the snapshots and add the new tests. Will merge when tests pass

🙈 I forgot that the snapshot tests depend on the title of the test

Issue #81, opened 2024-04-24
Handle `round_id`s other than dates?


annakrystalli:
### **Background**

Since the beginning of the project we have discussed and in general planned for supporting `round_id`s other than dates. So far this has not been necessary and our validations have focused on the assumption that `round_id`s will be dates, largely because all known hubs do indeed use dates as round IDs. Some discussing around this topic during development of the validation framework can be found here: https://github.com/Infectious-Disease-Modeling-Hubs/hubValidations/discussions/13

However, the[ push to convert historical hubs to hubverse style hubs has resurfaced the question ](https://github.com/orgs/Infectious-Disease-Modeling-Hubs/discussions/7#discussioncomment-9190990) of supporting non-date round ids and the need to assess implementation implications and weigh them against the benefits of supporting this feature.

### **Implications of using non-dates as round ids**

Use of a non-date round id has some important implications, most importantly on how submission windows are configured:
- When the round id is a date and is configured as a task ID contained in the file, multiple rounds can be concisely configured in a single round configuration by setting multiple options for a valid round id through the relevant task id as well as setting the submission window to be calculated as **[relative to the value of the round id](https://hubverse.io/en/latest/quickstart-hub-admin/tasks-config.html#step-8-set-up-submissions-due)**.
- If a character string is used instead as a round ID, the specification of a window relative to a date contained within the file (or filename) is no longer possible. This means that each round would need to be configured individually, with an explicitly set submission window `start` and `end` date for each round id. For example, if the following simplified flusight `tasks.json` config is changed to use epiweek (in the format `YYYY-EPIWEEK`) instead of ISO date, for just specifying 3 rounds, the size of the config triples containing primarily repeated information (only the submission window specification changes:
   #### **date round id**
    ```json
        {
        "schema_version": "https://raw.githubusercontent.com/Infectious-Disease-Modeling-Hubs/schemas/main/v2.0.0/tasks-schema.json",
        "rounds": [{
                "round_id_from_variable": true,
                "round_id": "origin_date",
                "model_tasks": [{
                    "task_ids": {
                        "origin_date": {
                            "required": null,
                            "optional": [
                                "2022-12-12", "2022-12-19", "2022-12-26"
                            ]
                        },
                        "target": {
                            "required": null,
                            "optional": ["wk ahead inc flu hosp"]
                        },
                        "horizon": {
                            "required": [2],
                            "optional": [1]
                        },
                        "location": {
                            "required": ["US"],
                            "optional": [
[...]

PR #133, opened 2024-10-24
Fix file name parse error (#132)

### LINKED ISSUE #132: space in filename doesn't throw validation error
See [this PR](https://github.com/reichlab/variant-nowcast-hub/pull/121) for details. A file was submitted that initially had a filename that had a whitespace before the `.parquet` extension. It passed validation. It should not have passed validation.

annakrystalli:
This PR resolves #132 by bolstering the parsing of file names in `parse_file_name()` to ensure they match the specific pattern required rather than just checking patterns of interest are extracted successfully.

I also added more fine-grained check with more informative error messages and fixed a bug in which the compression extensions were not being successfully recorded and returned.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 * Added:
   - new vignette on how to create custom validation checks for hub validations (#121) 
   - new section on how to manage additional dependencies required by custom validation functions (#22).
+* Bolstered parsing of file names in `parse_file_name()` to ensure they match the specific pattern required rather than just checking patterns of interest are extracted successfully and added more fine-grained check error messages  (#132).
 
 # hubValidations 0.7.0
 
--- a/R/parse_file_name.R
+++ b/R/parse_file_name.R
@@ -28,7 +28,6 @@
 parse_file_name <- function(file_path, file_type = c("model_output", "model_metadata")) {
   file_type <- rlang::arg_match(file_type)
   checkmate::assert_string(file_path)
-  file_name <- tools::file_path_sans_ext(basename(file_path))
 
   split_pattern <- stringr::regex(
     "([[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2})|[a-z_0-9]+",
@@ -39,10 +38,13 @@ parse_file_name <- function(file_path, file_type = c("model_output", "model_meta
     paste0(compress_codec, collapse = "|"),
     ")$"
   )
+  file_name <- tools::file_path_sans_ext(basename(file_path)) |>
+    # remove compression extension before validating and splitting
+    stringr::str_remove(compress_regex)
+
   split_res <- unlist(
     stringr::str_extract_all(
-      # remove compression extension before splitting
-      stringr::str_remove(file_name, compress_regex),
+      file_name,
       split_pattern
     )
   )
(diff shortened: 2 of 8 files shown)


nickreich (approved):
Thanks for the quick turn-around on this!

zkamvar (approved):
🏃🏽‍♀️⚡ THAT WAS FAST WORK! Splitting out the concerns of the valid structure and the extension, giving an informative response is 💯

annakrystalli:
Thanks for the reviews @nickreich and @zkamvar .

I'm going to merge but am wondering if we can put to be the following issue as well https://github.com/hubverse-org/hubValidations/issues/81

Offically this PR allows for round IDs that are not strictly ISO dates, so long as they match the pattern `[A-Za-z0-9_]`.  Overall the regex pattern for round IDs is:
`^(\\d{4}-\\d{2}-\\d{2})|[A-Za-z0-9_]+$`

Should I go ahead and open issues to:
- Add a regex pattern to the `tasks.json` `round_id` property
- Add a dynamic check to `valdiate_config()` that checks that `round_id`s match that pattern (if `tound_ids_from_variable` is `true`
- Explicitly mention this in a note in `hubDocs` in this section: https://hubverse.io/en/latest/user-guide/model-output.html#directory-structure

zkamvar:
> There is a test for `round_1` round ID here:
> 
> https://github.com/hubverse-org/hubValidations/blob/5c8395297181e2a2cd1f73661155511278d48438/tests/testthat/test-parse_file_name.R#L41-L50

Oh good point! Sorry about that. I'll close #81 as completed here.

nickreich:
Regarding the suggestion to
> Add a regex pattern to the `tasks.json` `round_id` property

I have forgotten where we are currently on this, but don't round IDs have to be `YYYY-MM-DD`s? I have a small worry about adding a complex regex expression as a field to a config file. Regex is a whole thing that is hard to understand for someone who hasn't interacted with them before. E.g. could we expect a hub admin to know what to put in here?

annakrystalli:
Totally agreed that regex are complex and confusing. Hence I'm suggesting a clarification in hubDocs that makes the expectation clear in plain language that doesn't require interpretating regex.

Having said that we have always accepted round IDs that are not ISO dates which is exactly why I'm wanting to formalise this so there aren't further misconceptions and having to answer this question over and over again.

nickreich:
Maybe we should migrate further discussion to a specific issue? I think basically I'm on board, and I like the idea of being more specific about round_id formatting. I guess as long as we provide several examples of commonly used formats for regex that hub admins could include in their config file it would be ok.

annakrystalli:
> Maybe we should migrate further discussion to a specific issue?

Can do but we've had multiple issues and discussions about this already:
- https://github.com/orgs/hubverse-org/discussions/10
- https://github.com/hubverse-org/hubValidations/issues/81
- https://github.com/hubverse-org/hubValidations/discussions/13
- https://github.com/orgs/hubverse-org/discussions/12

and we've established that the pattern above is what is accepted (and in truth has always been accepted. It just hasn't been documented well enough hence the lingering confusion about this).

What I'm proposing is to open specific issues that codify what has effectively always been the case  in the appropriate locations rather than open another discussion issue.

>  I guess as long as we provide several examples of commonly used formats for regex that hub admins could include in their config file it would be ok.

The hub admins would not need to provide a regex,. In the config it's the actual values that are encoded and the schema will perform the validation of the config, ensuring round id values match the expected pattern. Hence hub admins only need to ensure their proposed round IDs match the proposed pattern. Otherwise `validate_confi()` will complain (once we actually formally codify it).

We do already provide valid examples of what round IDs should look like in the schema here:
https://github.com/hubverse-org/schemas/blob/1b1b9b2e0810483c37faa386649e3083d5fcb05e/v3.0.1/tasks-schema.json#L28-L31

+ the idea is to bolster `hubDocs` documentation also,

annakrystalli:
Just to be crystal clear, hub admins just need to know that round IDs must be either:
- ISO dates (`YYYY-MM-DD`)
- Any other round identifier that contains letters, numbers of `_`

We can verbally explain this in `hubDocs` while the pattern suggested for the schema and dynamic validation of round ID values in the round ID column happens behing the scenes

PR #142, opened 2024-10-29
Update `exec_cfg_check` to use config source path relative to the hub root

### LINKED ISSUE #141: Custom checks fail if working directory is not hub root
brought up by @elray1 in https://github.com/reichlab/variant-nowcast-hub/issues/127

> ```
> > validate_submission(hub_path = '../variant-nowcast-hub',
> +                     file_path = 'UGA-multicast/2024-10-23-UGA-multicast.parquet') -> sub_validation
> Error in `purrr::map()`:
> ℹ In index: 1.
> ℹ With name: clade_prop_sum_one.
> Caused by error in `file()`:
> ! cannot open the connection
> Run `rlang::last_trace()` to see where the error occurred.
> Warning message:
> In file(filename, "r", encoding = encoding) :
>   cannot open file 'src/validations/R/clade_prop_sum_one.R': No such file or directory
> ```
> 
> He confirmed that validations run OK if he runs them from within the hub repo root, specifying `hub_path = '.'`.  It seems like somewhere, we should be targeting the `clade_prop_sum_one.R` file relative to the `hub_path`, but we're just using that relative path directly.

zkamvar:
There was a problem with `exec_cfg_check` where it was assuming that the working directory was the same directory as the hub directory. This PR fixes that issue by pulling `hub_path` from the caller's environment and using that to create a valid path. 

With this fix, however I had to also rewrite the tests because a lot of them were written in a way that worked very well to test that the execution order would be respected when the custom checks were called, but they did not test that the checks actually lived inside of the hub, which did pose a bit of a security concern because you effectively would end up with a situation where you could set an arbitrary URL as the source for the script: 

```
source: https://example.com/check.R
```

I added an extra test helper to do the following:

1. create a copy of a test hub
2. copy a check script to `src/validations/R` in the copy of the hub
3. create the `validations.yml` file with the correct parameters

This will fix #141


annakrystalli (commented):
Thanks for catching and fixing this! 

In all honesty I'm finding it harder to follow what's going on in the tests though.  I prefer having tangible config files that can be inspected easily online without having to pick through and figure out what the stand up code is doing.  

For me it would have been quicker and easier to follow if the deleted test yaml files were being used to configure each test instead of programmatically generating `validations.yml` files and the stand up function just took an argument that specified which `.yml` file to copy into the test hub. Then, once all contents of `testthat::test_path("testdata/src/R/")` were copied into the test hubs `src/validations/R/`) directory,  the `validations.yml` file could be used to configure which file to source (as we do in real hubs).

I think at least some comments to explain what going is needed in the stand up function.

zkamvar:
> For me it would have been quicker and easier to follow if the deleted test yaml files were being used to configure each test instead of programmatically generating validations.yml files and the stand up function just took an argument that specified which .yml file to copy into the test hub. Then, once all contents of testthat::test_path("testdata/src/R/") were copied into the test hubs src/validations/R/) directory, the validations.yml file could be used to configure which file to source (as we do in real hubs).

Understood. I was having a bear of a time with this. In short, I got confused and [hurt myself in my confusion](https://static.wikia.nocookie.net/all-fiction-battles/images/b/ba/Pokemon_status_effect.jpg/revision/latest?cb=20210108071251). I'll rejig.

annakrystalli:
No worries! BTW, could you add a note in `NEWS.md` too about the bug fix?

zkamvar:
> No worries! BTW, could you add a note in `NEWS.md` too about the bug fix?

Done! It's ready for another pass, @annakrystalli :smiley:

annakrystalli (approved):
Yeyy! Looks awesome and feels much easier to parse. Also ⚡️⚡️⚡️!

PR #144, opened 2024-10-30
Downgrade missing metadata file to `<check_failure>` when validation model output files. (#138)

### LINKED ISSUE #138: validating output files without an accompanying metadata file
If I try to validate a model output file without an accompanying metadata file, I get an early exit before most checks are performed:

```
> hubValidations::validate_submission(
+     hub_path = ".",
+     file_path = file.path("example-submission", paste0(reference_date, "-example-submission.csv")),
+     skip_submit_window_check = TRUE
+ )

── FluSight-forecast-hub ────

✔ [valid_config]: All hub config files are valid.

── 2024-11-16-example-submission.csv ────

✔ [file_exists]: File exists at path model-output/example-submission/2024-11-16-example-submission.csv.
✔ [file_name]: File name "2024-11-16-example-submission.csv" is valid.
✔ [file_location]: File directory name matches `model_id` metadata in file name.
✔ [round_id_valid]: `round_id` is valid.
✔ [file_format]: File is accepted hub format.
ⓧ [metadata_exists]: Metadata file does not exist at path model-metadata/example-submission.yml or
  model-metadata/example-submission.yaml.
```

I would think that most of the remaining checks would not depend on the existence of a model metadata file.  Is there any way to get around this?

annakrystalli:
Downgrade missing metadata file when checking model output submission to `<check_failure>` and remove early return. 

Resolves #138

### DIFF
--- a/R/check_submission_metadata_file_exists.R
+++ b/R/check_submission_metadata_file_exists.R
@@ -38,6 +38,6 @@ check_submission_metadata_file_exists <- function(file_path, hub_path = ".") {
     msg_subject = "Metadata file",
     msg_attribute = msg_attribute,
     msg_verbs = c("exists", "does not exist"),
-    error = TRUE
+    error = FALSE
   )
 }
--- a/R/validate_model_file.R
+++ b/R/validate_model_file.R
@@ -91,9 +91,6 @@ validate_model_file <- function(hub_path, file_path,
       file_path = file_path
     ), file_path
   )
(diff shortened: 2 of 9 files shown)


zkamvar (approved):
LGTM.

There is one caveat that if there are any custom metadata checks that require the metadata file, they will result in an `exec_error` and stop the validation process, but that's not a big issue since the `check_failure` will be there to indicate that the file does not exist.

elray1 (approved):

PR #145, opened 2024-10-30
Check that n of model output files per round per team sans ext does not exceed accepted n (#139)

### LINKED ISSUE #139: Error when a team submits both a csv and a parquet file for the same round
In hubs that accept multiple file types for model output submissions, a team could get passing validations if they submit both a csv file and a parquet file for the same modeling round.  This should result in a failing validation.

annakrystalli:
Best reviewed after #144 is merged

This PR checks that the number of model output files per round per team, regardless of extension, does not exceed accepted number of files (which defaults to 1L). The check returns a `check_failure` class object and does not lead to early return.

Resolves #139


elray1 (approved):
this looks good to me

PR #146, opened 2024-10-30
Allow for NA values in optional horizon time difference checks (#140)

### LINKED ISSUE #140: In optional validation for horizon time difference, allow for NA values
This season, the FluSight hub will include some standard weekly step-ahead targets, with columns `reference_date`, `horizon`, and `target_end_date`, where we expect that the `target_end_date` will be equal to the `reference_date` plus `horizon` weeks.  However, the hub will also accept some other targets that are for seasonal quantities like peak incidence and peak timing, for which the `horizon` and `target_end_date` columns are not relevant, and `NA` values will be supplied.  In that instance, the optional validation for consistency between the `reference_date`, `horizon`, and `target_end_date` columns fails in the rows with NA values:

```
[horizon_timediff]: missing value where TRUE/FALSE needed
```

I think maybe we need to update these lines to only check the rows where `compare` is non-missing (trusting to other validation steps to ensure that the relevant values are missing only for the appropriate target groups).

https://github.com/hubverse-org/hubValidations/blob/5c8395297181e2a2cd1f73661155511278d48438/R/opt_check_tbl_col_timediff.R#L44-L45

https://github.com/hubverse-org/hubValidations/blob/5c8395297181e2a2cd1f73661155511278d48438/R/opt_check_tbl_horizon_timediff.R#L54-L55

annakrystalli:
Should be reviewed after #144 & #145 are merged

This PR allows for NA values in optional horizon time difference checks by subsetting `tbl` to complete cases of relevant columns. This ensures rows that may not be targeting relevant to modeling task do not cause false check failure.

Resolves #140


zkamvar (approved):
LGTM!

TIL about `stats::complete.cases()`! 

As an aside, is there a reason why the content from #144 and #145 are in here?

annakrystalli:
> As an aside, is there a reason why the content from https://github.com/hubverse-org/hubValidations/pull/144 and https://github.com/hubverse-org/hubValidations/pull/145 are in here?

Not sure. The diffs should have updated automatically when the other PR's were merged in. That's how it always worked before. GitHub has been a bit strange recently, like when it closed the PR when the branch it was pointing to was deleted. I'm, 100% sure in the past it just redirected the PR to `main` in such a situation.

zkamvar:
> Not sure. The diffs should have updated automatically when the other PR's were merged in. That's how it always worked before. GitHub has been a bit strange recently, like when it closed the PR when the branch it was pointing to was deleted. I'm, 100% sure in the past it just redirected the PR to main in such a situation.

But was this pattern you were observing based on the [before R-universe workflow](https://hubverse-org.github.io/hubDevs/articles/release-process.html#before)? In that workflow, there were always a chain of pull requests `main <- A <- B <- C` and they would be merged in the order of C, B, and finally A. When GitHub deleted the branch last time, it was because the same chain existed, but they were merged in the order of A, B, and finally C. When you merged A, the branch that B was pointing to no longer existed, so the pull request was closed. 

In this case, you had A, B, and C all pointing to the `main` branch, but A, B, and C were still dependent on each other (because they were created sequentially). 

I would recommend [creating each new branch from main](https://hubverse-org.github.io/hubDevs/articles/release-checklists.html#subsequent-pr-checklist) in the future.

annakrystalli:
Sometimes I need the features in the other PRs too.

It has worked in the past and when I need to I will still use this approach.

annakrystalli:
What is actually needed in this situation seems to be a merge https://github.com/isaacs/github/issues/750#issuecomment-531642134

PR #150, opened 2024-10-31
refactor custom function execution

### LINKED ISSUE #148: Refactor exec_cfg_check.R
As we now have hubs that are using custom functions, this is a good opportunity to refactor the `exec_cfg_check.R` file a bit. I ran into difficulty working with this in #142 partially because the file has a lot of abbreviations and no explanatory comments.

I would like to refactor this to add some explanatory comments and refactor some parts to avoid repetition and use boolean sentinel variables. For example, we could refactor this: 

https://github.com/hubverse-org/hubValidations/blob/5c8395297181e2a2cd1f73661155511278d48438/R/exec_cfg_check.R#L36-L42

to use sentinel variables and comments

```r
get_caller_env_formals <- function(fn, caller_env, cfg_args) {
  # find the arguments of the custom function
  fn_arg_names <- rlang::fn_fmls_names(fn)
  # variables available from the calling environment (e.g. hub_path, tbl, etc...)
  available_vars <- rlang::env_names(caller_env)
  # match the arguments from the calling environment, discarding the ones not needed
  args_from_vars <- fn_arg_names %in% available_vars
  # do not include arguments that are specified in the config. 
  not_config_args <- !fn_arg_names %in% names(cfg_args)
  caller_env_fmls <- fn_arg_names[args_from_vars & not_config_args]
  # extract these values from the calling environment, replacing with `NULL` if they are missing.
  rlang::env_get_list(caller_env, nms = caller_env_fmls, default = NULL)
}
```

I would be happy to tackle this.

zkamvar:
I've refactored the custom function execution functions. This refactor is mainly for the purposes of debugging and readability. I find this especially important when navigating a function that works with variables in other environments. Having the comments and the sentinel variables reduces the cognative load necessary to walk through the function during a debugging session. 

During my refactor, I found that there was an untested error and a dangling else condition that should have resulted in an error in the case of a malformed config file. I added classes to those errors and added tests for them.

Note that this PR does not need to ship with the next version of hubValidations if we want to ship the new features immediately.

This will fix #148

### DIFF
--- a/R/exec_cfg_check.R
+++ b/R/exec_cfg_check.R
@@ -1,7 +1,7 @@
 #' Execute a check function from the validations configuration file
 #'
 #' @param check_name [character] the name of the check function
-#' @param validations_cfg [list] the the parsed `validaitons.yml` file
+#' @param validations_cfg [list] the the parsed `validations.yml` file
 #' @param caller_env [environment] the environment of the calling function.
 #'   This is usually generated from `rlang::caller_env()`
 #' @param caller_call [call] the call of the calling function.
@@ -19,8 +19,8 @@ exec_cfg_check <- function(check_name, validations_cfg, caller_env, caller_call)
     )
   } else if (from_src) {
     # TODO: Validate source script.
-    # if it's a source script, we need to run the script locally to extract
-    # the function.
+    # if it's a source script, we need to source the script locally to make
+    # the function available in local environment.
     hub_path <- rlang::env_get(env = caller_env, nm = "hub_path")
     src <- fs::path(hub_path, fn_cfg[["source"]])
     source(src, local = TRUE)
--- a/R/exec_cfg_check.R
+++ b/R/exec_cfg_check.R
@@ -1,25 +1,40 @@
+#' Execute a check function from the validations configuration file
+#' 
+#' @param check_name [character] the name of the check function
+#' @param validations_cfg [list] the the parsed `validaitons.yml` file
+#' @param caller_env [environment] the environment of the calling function.
+#'   This is usually generated from `rlang::caller_env()`
+#' @param caller_call [call] the call of the calling function.
+#'   This is usually generated from `rlang::caller_call()`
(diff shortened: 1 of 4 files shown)


R/exec_cfg_check.R
```diff
@@ -1,25 +1,48 @@
+#' Execute a check function from the validations configuration file
+#'
+#' @param check_name [character] the name of the check function
+#' @param validations_cfg [list] the the parsed `validaitons.yml` file
```
annakrystalli, line 4:
```suggestion
#' @param validations_cfg [list] the the parsed `validations.yml` file
```

R/exec_cfg_check.R
```diff
@@ -1,25 +1,48 @@
...
+    # if the function is from a package, assume the package is installed and
+    # extract it from that package.
     fn <- get(fn_cfg[["fn"]],
       envir = getNamespace(fn_cfg[["pkg"]])
     )
-  } else if (!is.null(fn_cfg[["source"]])) {
-    # TODO Validate source script.
+  } else if (from_src) {
+    # TODO: Validate source script.
+    # if it's a source script, we need to run the script locally to extract
```
annakrystalli, line 22:
```suggestion
    # if it's a source script, we need to source the script locally to make
```

R/exec_cfg_check.R
```diff
@@ -1,25 +1,48 @@
...
+    # the function.
```
annakrystalli, line 23:
```suggestion
    # the function available in local environment.
```

annakrystalli (approved):
Thanks for this refactor! 

I have to be honest, I feel the level of commenting is a bit much. But I like the extra tests and checks and some of the sentinel variables do make it easier to follow.

 Overall definitely an improvement in robustness and readability 👍

R/exec_cfg_check.R
zkamvar: 🙈

Issue #165, opened 2024-11-20
The main argument in `submission_tmpl()` should be `hub_path`. `hub_con` and `config_tasks` should be deprecated


annakrystalli:
Currently there is an option to provide either a hub_connection class object to `hub_con`, which has `config_tasks` attached as an attribute) or a `config` class object to `config_tasks`. It feels easier to just ask for `hub_path` instead of requiring folks to use `hubData::connect_hub()` to create a `hub_connection` given we only need the actual `config_tasks` object from the connection.

I would suggest we just do away with `config_tasks` (which also requires an extra step to read the config) and restrict the argument to `hub_path` only, then use `read_config()` internally to get the config.

This would also resolve #137

PR #167, opened 2024-11-21
Read derived taskids from config (#155)

### LINKED ISSUE #155: Read derived_task_ids from config
Allow for derived_task_ids to be read from config: follow `output_type_id_datatype` paradigm where the default behaviour is to get the value from the config but can also be overriden through the `derived_task_ids()`

annakrystalli:
This PR resolves #155 by ensured derived task IDs are read from config files by default. I've included updates to the relevant vignettes.

There were also a couple more issues that needed fixing before tests could pass.

1. To fix a bug with validating v4, there was some superseded code in `check_tbl_values_required()`  I needed to remove. This code was required initially when validation was being performed on the submission file as a whole. During the last optimisation round, validation was split across modeling tasks, with each subset of the file having it's own subset of the full and required value grids. Because of this, the code being deleted is not required anymore because it was dealing with complications from the fact that originally data from different modeling tasks were being considered together and you could end up with a required grid with values in rows of a column and `NA`s in other rows which related to a different modeling task. This is no longer the case and the code could have previously be removed for efficiency but wasn't causing issues. However as a result of changes in v4 it had in fact turned into a bug and was causing false positives to be being returned and hence needed removing. All tests still pass after removal.
2. The `mockery::stub` functions was leaking it's stub into subsequent tests! This was so frustrating to track down that I went ahead and removed all instances of the function from our tests and the package from our dependencies, replacing them with `local_mocked_bindings()`. Because this only works on functions in a package namespace, I needed to ensure mocked functions were imported formally into the namespace and also to add an internal binding for `Sys.time` before I could mock it. This proved a good opportunity though to go through and tidy up by ensuring we are using the re-exported versions of functions internally.

Ideally I would not have mixed these in to this PR but given I needed to to fix broken tests, I went ahead and included them. As such, it might be useful to look at individual commits.


R/check_tbl_values_required.R
```diff
@@ -145,22 +145,7 @@ check_modeling_task_values_required <- function(tbl, req, full) {
 # values of successively smaller n.
 missing_required <- function(x, mask, req, full) {
   opt_cols_list <- get_opt_col_list(x, mask, full, req)
-
```
annakrystalli, line 148: The following code is the buggy code that needed removing

R/check_tbl_values_required.R
```diff
@@ -187,32 +172,21 @@ get_opt_col_list <- function(x, mask, full, req) {
 
 # Identify missing required values for optional value combinations.
 # Output full missing rows compiled from optional values and missing required values.
```
annakrystalli, line 174: Same below

R/validate_model_data.R
```diff
@@ -112,6 +116,19 @@ validate_model_data <- function(hub_path, file_path, round_id_col = NULL,
     return(checks)
   }
```
annakrystalli, line 118: Adding the extraction of `derived_task_ids` here so that it follows checks on `round_id`

tests/testthat/_snaps/validate_submission.md
```diff
@@ -1219,7 +1219,7 @@
     Code
       str(dup_model_out_val)
     Output
-      Classes 'hub_validations', 'list'  hidden list of 10
```
annakrystalli, line 1222: These snapshots were incorrectly picking up the error from the leaky stub. Now corrected

zkamvar (approved):
Thank you for this and for the explanations! 

I think this makes sense and I'm curious about one thing. (NOTE: this is me thinking out loud about the process, not a criticism) It looks like `submission_tmpl()` and `validate_model_data()` have a catch for a `NULL` `derived_task_ids` argument to populate the task IDs specified in v4 hub task configuration files (earlier hubs will continue to return `NULL`) while lower-level functions (i.e. `expand_model_out_grid()`) do not have this catch.

This means that (in v4 hubs with non-null `derived_task_ids`) when validating a submission or creating a submission template, it's possible to validate with a _subset_ of derived task IDs ignored, but it's not possible to validate with all of them ignored (because `NULL` will default to the defined `derived_task_ids` property). This makes sense to me because the whole point of having this property is to avoid errors from invalid combinations of derived task IDs and the task IDs they depend on. 

This is great work and well done on the detective work with mockery!

annakrystalli:
> This means that (in v4 hubs with non-null `derived_task_ids`) when validating a submission or creating a submission template, it's possible to validate with a _subset_ of derived task IDs ignored, but it's not possible to validate with all of them ignored (because `NULL` will default to the defined `derived_task_ids` property). This makes sense to me because the whole point of having this property is to avoid errors from invalid combinations of derived task IDs and the task IDs they depend on.

Indeed. Well spotted. There are also practical reasons for this:
1. In `validate_model_data()` `round_id`s are not provided as arguments, instead being parsed from the file name. So we need that step to have completed successfully before we can extract `derived_task_ids` (as round level `derived_task_ids` are allowed, hence we need to know the `round_id` first.
2. In `submission_tmpl()` there are 2 potential main inputs, `hub_con` or `config_tasks`. We need to know which one of these is supplied before we can determine where `config_tasks` will be coming from. Hence it is not possible to set a function as the input to `derived_taks_ids`. This might change though if we implement #165

PR #175, opened 2024-12-11
Surface force_output_types in submission_tmpl. Resolves #162

### LINKED ISSUE #162: Surface `force_output_types` in `submission_tmpl()`
When creating a submission template for a v4 configured, to enable including the output type ID values of an optional output types a team might want to submit to, `force_output_types` will need to be used in underlying `expand_model_out_grid()`. To enable this we need:
- [x] surface the `force_output_types` argument in `submission_tmpl()`
- [x] document its use in roxygen.
- [ ] #164

annakrystalli:
Tis PR surfaces the `force_output_types` argument in the `submission_tmpl()` function, allowing users to force optional output types to be included in a submission template when `required_vals_only = TRUE`. In conjunction with the use of the `output_types` argument, this allows users to create submission templates which include optional output types they plan to include in their submission.


annakrystalli:
Thanks @zkamvar and well spotted! I committed all your suggestions and tests all pass.

zkamvar (approved):
Thank you!

PR #176, opened 2024-12-11
Add `check_tbl_derived_task_id_vals()` function

### LINKED ISSUE #110: Add simple value checks for derived task IDs
In #109 we introduced the ability to ignore derived task IDs during validation to improve validation performance.

This effectively encodes such task IDs as NAs during relevant checks.

While [administrators are advised in the documentation to ensure custom checks guarantee the validity of derived task ids ](https://github.com/hubverse-org/hubValidations/blob/8ec826172e19d4953c25f01583a311bc5157a99a/vignettes/articles/validate-pr.Rmd#L77-L79) with respect to their relationship to the task IDs they are derived from, it would also be good to add a simple check to ensure that derived task ID at least match valid values in specified in the config (as opposed to combinations with other task ID values etc).

This should be simple and to taxing computationally.

annakrystalli:
This PR adds the `check_tbl_derived_task_id_vals()` check to `validate_model_data()` that ensures that values in derived task ID columns match expected values for the corresponding derived task IDs in the round as defined in `tasks.json` config Given the dependence of derived task IDs on the values of other values, the check ignores the combinations of derived task ID values with those of other task IDs and focuses only on identifying values that do not match corresponding accepted values. This ensures `derived_task_ids` values which are ignored in all other standard value checks receive atl east some basic validation.

Resolves #110


tests/testthat/test-validate_submission.R
```diff
@@ -327,20 +327,6 @@ test_that("Ignoring derived_task_ids in validate_submission works", {
...
-      file_path = "flu-base/2022-10-22-flu-base.csv",
-      skip_submit_window_check = TRUE,
-      derived_task_ids = "target_end_date"
-    ),
-    validate_submission(
-      hub_path = system.file("testhubs/samples", package = "hubValidations"),
-      file_path = "flu-base/2022-10-22-flu-base.csv",
-      skip_submit_window_check = TRUE
-    )
-  )
```
annakrystalli, line 343: Removed this because of the introduction of the new check, now when `derived_task_ids` are NULL the check is skipped and a message returned whereas if `derived_task_ids` are included, the result of the check is retuned. Hence these two situations do not return the same result anymore

R/check_tbl_derived_task_id_vals.R
```diff
@@ -0,0 +1,92 @@
...
+    tbl[derived_task_ids], derived_task_id_vals,
+    \(.x, .y) {
+      setdiff(.x, .y)
+    }
+  )
+
+  invalid_derived_task_ids <- purrr::map_lgl(
+    setdiff_vals,
+    \(.x) length(.x) > 0L
+  )
```
zkamvar, line 52: This is equivalent and easier to read. 

```suggestion
  invalid_derived_task_ids <- lengths(setdiff_vals) > 0L
```

R/check_tbl_derived_task_id_vals.R
```diff
@@ -0,0 +1,92 @@
...
+    config_tasks = config_tasks,
+    round_id = round_id,
+    derived_task_ids = NULL
+  )[derived_task_ids]
+
+  setdiff_vals <- purrr::map2(
+    tbl[derived_task_ids], derived_task_id_vals,
+    \(.x, .y) {
+      setdiff(.x, .y)
+    }
```
zkamvar, line 46: Without extra arguments, there is no need for a wrapper function. 

```suggestion
    tbl[derived_task_ids], derived_task_id_vals, setdiff
```

zkamvar (approved):
This looks good to me! There are two simplifications to the code that I suggested.

annakrystalli:
Thanks @zkamvar ! Feel free to push forward with suggestions and merge

PR #178, opened 2024-12-12
:bug: Separate v4 req vals checks by output type. Resolves #177

### LINKED ISSUE #177: unexpected `req_vals` failure in v4 hub
I don't quite understand the shape of the problem yet, so it's difficult to provide a reprex.

A problem occurred in https://github.com/cdcepi/FluSight-forecast-hub/pull/1261 where the baseline model was submitted with samples for four of the five optional horizons. This particular model failed because samples were not included for one horizon (see https://github.com/cdcepi/FluSight-forecast-hub/pull/1261#issuecomment-2537144503):

```r
library(gert)
#> Linking to libgit2 v1.8.1, ssh support: YES
#> Global config: /Users/zkamvar/.gitconfig
#> Default user: Zhian N. Kamvar (UMass) <zkamvar@umass.edu>
library(withr)
library(hubValidations)

tmp <- tempfile()
git_clone("https://github.com/cdcepi/FluSight-forecast-hub.git", tmp)
with_dir(tmp, system("gh pr checkout 1261"))
res <- with_dir(tmp, validate_pr(gh_repo = "cdcepi/FluSight-forecast-hub", pr_number = 1261))

options(width = 120)
print(res$req_vals$missing, n = Inf)
#> # A tibble: 53 × 7
#>    reference_date horizon target          target_end_date location output_type output_type_id
#>    <date>           <int> <chr>           <date>          <chr>    <chr>       <chr>         
#>  1 2024-12-14          -1 wk inc flu hosp NA              01       sample      <NA>          
#>    ... truncating because only location changes ...         
#> 53 2024-12-14          -1 wk inc flu hosp NA              US       sample      <NA>
```

This is perplexing because [samples are not required in this hub](https://github.com/cdcepi/FluSight-forecast-hub/blob/b6b42de41839eef2b407581e9210e384946bc63b/hub-config/tasks.json#L306-L319), and they do not reference "horizon":

```json
                        "sample": {
                            "output_type_id_params": {
                                "type": "integer",
                                "min_samples_per_task": 100,
                                "max_samples_per_task": 100,
                                "compound_taskid_set" : ["reference_date", "location", "target"]
                            },
                            "value": {
                                "type": "integer",
                                "minimum": 0
                            },
                            "is_required": false
                        }
                    },
```

As suggested by @elray1, I tested this using the v3 version (see https://github.com/cdcepi/FluSight-forecast-hub/pull/1261#issuecomment-2537242929) and it was successful, so we reverted the hub back to v3.

annakrystalli:
This PR fixes the bug reported in #177 . It appeared that the new logic behind optional output types having all required values was not being handled properly and as a result evaluations of different output types were leaking into checks of other output types returning non-sensical false positives.

Given the fact that in v4 we have a much easier way of determining whether an output type should be examined (all `is_required` + any optional output type submitted, an optimisation I was planning to make to reduce memory pressure and speed up required value validation was to chunk the evaluation by output type, sth that was not possible with v3. Turns out this optimisation also keeps things clean and fixes the bug reported in #177! Double win!


zkamvar (approved):
Thank you for the quick turnaround on this! The solution (checking each output type individually) makes sense and the test is clear. And I like that you included the issue number in the tests. 

At some point I'd like to pick your brain about how these checks work and why this approach can only work for v4, but that's for later.

PR #188, opened 2025-01-07
add failing test for check of ascending columns

elray1:
This PR adds a currently-failing test case for the check of ascending values.  The test hub has two task groups with different quantile levels requested: The first asks for [quantiles at levels 0.1, 0.5, and 0.9](https://github.com/hubverse-org/hubValidations/blob/d2517dd581d11c306ae3cfe91e775f40ad48652c/tests/testthat/testdata/hub-diff-otid-per-task/hub-config/tasks.json#L84-L86), while the second asks for [quantiles at 0.1, 0.2, ..., 0.9](https://github.com/hubverse-org/hubValidations/blob/d2517dd581d11c306ae3cfe91e775f40ad48652c/tests/testthat/testdata/hub-diff-otid-per-task/hub-config/tasks.json#L197-L205).  A test submission file with correct ordering fails because its quantiles are re-ordered to the levels [0.1, 0.5, 0.9, 0.2, 0.3, 0.4, 0.6, 0.7, 0.8]


elray1:
I thought that a solution might be to update `order_output_type_ids` as follows:

```
order_output_type_ids <- function(tbl, config, types = c("cdf", "quantile")) {
  # step 1: create a lookup table from the config
  cdf_and_quantile <- config$output_type %in% types
  order_ref <- config[cdf_and_quantile, , drop = FALSE]
  # step 2: join
  tbl$output_type_id <- as.character(tbl$output_type_id)
  tbl_cols <- colnames(tbl)
  join_cols <- tbl_cols[tbl_cols != "value"]
  dplyr::inner_join(order_ref, tbl, by = join_cols)
}
```

However, this led to errors like:
```
Error in `dplyr::inner_join(order_ref, tbl, by = join_cols)`: Can't join `x$origin_date` with `y$origin_date` due to incompatible types.
i `x$origin_date` is a <character>.
i `y$origin_date` is a <date>.
```

tests/testthat/test-check_tbl_value_col_ascending.R
```diff
@@ -163,3 +163,22 @@ test_that("(#78) check_tbl_value_col_ascending will sort even if the data doesn'
...
+  tbl <- arrow::read_csv_arrow(
+    fs::path(hub_path, "model-output/2024-01-10-ILI-model.csv")
+  ) %>%
+    hubData::coerce_to_character()
+  file_path <- "ISI-NotOrdered/2024-01-10-ILI-model.csv"
+  file_meta <- parse_file_name(file_path)
+  expect_s3_class(
+    check_tbl_value_col_ascending(tbl, file_path, hub_path, file_meta$round_id),
+    c("check_success", "hub_check", "rlang_message", "message", "condition"),
+    exact = TRUE
```
zkamvar, line 179: I appreciate checking the class instead of adding another snapshot test. Removing the extra classes that we don't control will make this test less flaky. 

```suggestion
    "check_success"
```

tests/testthat/test-check_tbl_value_col_ascending.R
```diff
@@ -163,3 +163,22 @@ test_that("(#78) check_tbl_value_col_ascending will sort even if the data doesn'
...
   expect_equal(actual, expected, ignore_attr = TRUE)
 })
+
+
+test_that("check_tbl_value_col_ascending works when output type IDs differ by target", {
+  hub_path <- test_path("testdata/hub-diff-otid-per-task/")
+  tbl <- arrow::read_csv_arrow(
+    fs::path(hub_path, "model-output/2024-01-10-ILI-model.csv")
+  ) %>%
+    hubData::coerce_to_character()
```
zkamvar, line 173: shouldn't this be `hubData::connect_model_output()` followed by `hubData::collect_hub()`?

elray1: I'm honestly not sure.  I have the sense that maybe not, because the connect/collect workflow is only guaranteed to work after the model output file has passed validations, and at this point we're still in the midst of those validations.  But on the other hand, using connect/collect might work just fine after some basic/preliminary checks have passed?

elray1: Maybe the best thing to do would be to replicate the way it's read in during running validations, which is to use [`read_model_out_file`](https://github.com/hubverse-org/hubValidations/blob/353e8db432ecf03f093d595e116d377400d1235b/R/validate_model_data.R#L67)?

zkamvar: Ah... it's actually `hubValidations::read_model_out_file()`

zkamvar: once again, I need to refresh before commenting 🙈

PR #190, opened 2025-01-10
refactor check_tbl_col_ascending

### LINKED ISSUE #189: refactor check_tbl_vals_cols_ascending(): add derived task IDs and improve performance
This was brought up from #105:

> Sorry! My brain was fried by arrow earlier and I didn't think through the performance implications of creating one expanded grid for both output types and all model tasks in one go. If we find this now slows some validation workflows down we could just perform the check one output type at a time. Sth to monitor.
>
> More importantly we should definitely ignore derived task IDs in both the expanded grid and `tbl` asap as this can have a real impact on performance.
> 
> @zkamvar would you be able to open a couple of issues for these? And ideally address the derived IDs issue asap? There's a few validation functions where something similar is happening, eg checking that all values are valid off the top of my head so it should be pretty straightforward to implement.
>
> _Originally posted by @annakrystalli in https://github.com/hubverse-org/hubValidations/issues/105#issuecomment-2578548335_

zkamvar:
In #105, I had implemented a sorting method for the `check_tbl_col_ascending()` function that involved using an expanded model grid subset to the `cdf` and `quantile` output types to sort the submission table. As pointed out by @annakrystalli, there were two problems:

1. I had neglected to pass `derived_task_ids` to the function
2. my implementation was inefficient because I used `bind_model_tasks = TRUE`

While I was looking at this for refactoring, I found that I had re-invented the function to sort the table by output type ID and the task IDs: `match_tbl_to_model_task()`.

Happily, this speeds up the process quite a bit _and_ allows me to delete code!

This will fix #189 

EDIT: huge thanks to @annakrystalli for implementing the correct solution and better tests for derived task IDs


annakrystalli:
Thanks for implementing this so quickly!

Instead of a review, as there were quite a few changes I wanted to propose, I went ahed and made a PR instead here: #191

annakrystalli (changes requested):
See #191

PR #191, opened 2025-01-13
Ak/check derived task ids/189

annakrystalli:
Instead of reviewing I've just gone ahead and made a PR with my suggestions.

This PR build on #190 but introduces the following changes:
- Instead of converting all but the value column to character, it expects an all character `tbl` and converts just the value column to numeric. This is how the original test was setup and the conversion of `value` to numeric was already happening in `check_values_ascending()` on the subset of relevant rows for the check. 
- I've updated the function documentation to reflect the above input change.
- To avoid unnecessary computations, I've taken the output types to check from the tbl itself instead of the config. That way we do not spend time creating expanded grids for optional output types in the config that do not exist in the data.
- I also went ahead and split the check by output type for maximum memory efficiency.
- I reverted the input to the check in `validate_model_data()` to be the `tbl_chr` version of the data. This exists in the function environment as a number of checks depend on the all character version of the model output data.
- I corrected the tests to reflect this input expectation. I also removed unnecessary `hubValidations` namespace specifications
- I added a test that checks derived task IDs are indeed ignored.

I've tested locally and all tests pass but will be good to run the tests in your PR once merged in before merging into `main`


R/match_tbl_to_model_task.R
```diff
@@ -36,7 +36,7 @@ match_tbl_to_model_task <- function(tbl, config_tasks, round_id,
     config_tasks,
     round_id = round_id,
     required_vals_only = FALSE,
-    all_character = TRUE,
+    all_character = all_character,
```
annakrystalli, line 39: I realised this argument wasn't being passed so fixed it

R/check_tbl_value_col_ascending.R
```diff
@@ -70,9 +67,34 @@ check_tbl_value_col_ascending <- function(tbl, file_path, hub_path, round_id,
...
+    tbl,
+    config_tasks = config_tasks,
+    round_id = round_id,
+    output_types = output_type,
+    derived_task_ids = derived_task_ids
+  ) %>%
+    purrr::compact()
+
+  purrr::map(output_type_tbls, check_values_ascending) %>%
+    purrr::list_rbind()
```
zkamvar, line 92: Suggestion to make the variable name a bit more clear and remove the `map` and `list_rbind` that are not needed. 

```suggestion
  sorted_tbl <- match_tbl_to_model_task(
    tbl,
    config_tasks = config_tasks,
    round_id = round_id,
    output_types = output_type,
    derived_task_ids = derived_task_ids
  ) %>%
    purrr::compact() %>%
    purrr::pluck(1)
  check_values_ascending(sorted_tbl)
```

My reasoning: I didn't quite understand why `compact()`, `map()`, and `list_rbind()` were used here if we are testing a single output type. I went down a rabbit hole trying to figure out why I couldn't just pass the first item of output from `match_tbl_to_model_task()` to `check_values_ascending()` since we are subsetting by a single output type. But it turns out that the output of `match_tbl_to_model_task()` (and, subsequently, `expand_model_out_grid()`) returns one list item per output type.


Instead of using `map()` and `list_rbind()` to loop over a one-element list, I added `purrr::pluck(1)` to get the data frame and passed the value to `check_values_ascending()`.

zkamvar (changes requested):
Thank you for the improvements, Anna! It all looks good with two things that can be tweaked:

1. the value column still needs to be converted to numeric
2. replacing a map over a single-element list with calling the function directly (see comment)

Additionally, should the function signature for `check_tbl_value_col_ascending()` be `tbl_chr` instead of `tbl` so that it's clear it expects a character-only tbl?

R/check_tbl_value_col_ascending.R
```diff
@@ -27,28 +27,25 @@ check_tbl_value_col_ascending <- function(tbl, file_path, hub_path, round_id,
   }
```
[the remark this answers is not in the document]
annakrystalli: It happens on line 105 https://github.com/hubverse-org/hubValidations/blob/018d024cdf1bb3f369fa0a31ab3011c05f43bf90/R/check_tbl_value_col_ascending.R#L105

This means we only convert the `value` column after subsetting (not the full `tbl` column)

R/check_tbl_value_col_ascending.R
annakrystalli: `match_tbl_to_model_task()` outputs **a list element per model task** not output type. If more than one output type is requested rows for all output types are included in each model task element.

We need compact to get rid of any `NULL` list elements which occur when an output type is not present in the config for a given model task. The code in this function is exactly as it appeared before in the main function, just wrapped in a new function so we can iterate over output types one at a type (which results in smaller outputs of `match_tbl_to_model_task()` at each iteration.

Therefore this suggestion won't work as you'd effectively only be processing the first model task.

R/check_tbl_value_col_ascending.R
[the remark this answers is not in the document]
zkamvar: Good catch and good point! 🙈

R/check_tbl_value_col_ascending.R
zkamvar: OHHHHHHH that's why! 

So the reason why the examples and tests only show one output type per element is because each model task has a separate output type? The reason it shows two elements is because there are two model tasks per round. 

If there were a hub that had two separate model tasks (e.g. for different targets) that were both providing quantile estimates, then `match_tbl_to_model_task()` would return two data frames when using `output_type = "quantile"`.

zkamvar: Thank you for being patient with me on this.

PR #193, opened 2025-01-22
Shorten message in check_tbl_value_col_ascending. Resolves #143

### LINKED ISSUE #143: Consider re-wording non-decreasing value column check message
Currently the message printed for the non-descending value column check is quite long:

```
✔ [value_col_non_desc]: Values in `value` column are non-decreasing as output_type_ids increase for all unique task ID value/output type combinations of quantile or cdf output types
```

It would be clearer if we moved the part about the check only being for quantile and cdf output types to the beginning, as well as prevent misunderstandings if the last part of the message is cut off (or people don't read that far)

annakrystalli:


lshandross (approved):

zkamvar (approved):
This makes it much more clear! Thank you!

Issue #203, opened 2025-03-10
Missing required target not caught by `check_tbl_values_required()`


LucieContamin:
I have an input table with only quantile output for one optional target (here `cum hosp`) and missing the required sample output for the required target (here `inc hosp`). I tested the table with `check_tbl_values_required()` and expected an error but the function did not return any issue and I am not sure if I miss something or if there is a bug somewhere. 

```r
df <- arrow::read_parquet("../exp/model-output/team4-modeld/2023-11-12-team4-modeld.parquet")
tbl_chr <- hubData::coerce_to_character(df)
table(df$target, df$output_type)
          
 #>           quantile
 #>  cum hosp     6900
hubValidations::check_tbl_values_required(tbl_chr, round_id = "2023-11-12", file_path = "2023-11-12-team4-modeld.parquet", hub_path = "../exp", derived_task_ids = NULL)
#> <message/check_success>
#> Message:
#> Required task ID/output type/output type ID combinations all present.
```
and here the required part of the `tasks.json` file:

```json
"model_tasks":[
				{
					"task_ids":
					{
						"origin_date":{
							"required":["2023-11-12"],
							"optional":null
						},
						"scenario_id":{
							"required":["A-2023-10-27","B-2023-10-27","C-2023-10-27","D-2023-10-27","E-2023-10-27"],
							"optional": null
						},
						"target":{
							"required":["inc hosp"],
							"optional":null
						},
						"horizon":{
							"required":[1,2,3,4,5,6,7,8,9,10],
							"optional":null
						},"location":{
							"required":null,
							"optional":["US","06","08"]
						},"age_group":{
							"required":["65-130","0-130"],
							"optional":["5-17"]
						}
					},
					"output_type":{
						"sample":{
							"output_type_id_params": {
								"type": "integer",
							    "min_samples_per_task": 100,
							    "max_samples_per_task": 200,
							    "compound_taskid_set": ["origin_date", "scenario_id", "target", "location"]
							},
							"value":{
								"type":"double",
								"minimum":0
							},
							"is_required": true
						}
					}, 
```

annakrystalli:
Hey @LucieContamin ! 

[...]

PR #204, opened 2025-03-12
Ignore derived task IDs when assessing model tasks for missing required values 

annakrystalli:
This PR does not resolve but is related to #203 which prompted uncovering this exception.

Evaluation of model output data for required values happens on data separated into model tasks. If all task IDs in a given model task have at least one required value, then some data needs to be submitted to this model task. This means the check should throw an error if data for such a model task is missing all together. If any task Ids are all optional though, it is not a requirement that that particular model task is submitted to. However, derived task Ids should be exempt from this check as they will always be all optional but result from the values of other task IDs. So only non-derived task IDs should be used to assess whether a model task is required or not.

This PR enables handling this exception.

It also contains a fix for explicit return linter errors.


tests/testthat/testdata/configs/tasks-v5-req-task-id.json
```diff
@@ -0,0 +1,171 @@
...
+                            },
+                            "target_type": "ordinal",
+                            "description": "This target represents a categorical severity level for rate of new hospitalizations per week for the week ending [horizon] weeks after the reference_date, on target_end_date.",
+                            "is_step_ahead": true,
+                            "time_unit": "week"
+                        }
+                    ]
+                },
+                {
+                    "task_ids": {
```
annakrystalli, line 81: Note that in this model task, all task IDs apart from the derived task ID `target_end_date` have a required value. The model task is completely missing however from the parquet test file below though.

zkamvar (approved):
This looks good! I appreciate the detailed explanation in the function comments.

annakrystalli:
Thanks @zkamvar !

PR #212, opened 2025-03-28
use title case

zkamvar:
This is a test commit to check the status of the Netlify preview


zkamvar:
The pkgdown preview works, so I will close this.

PR #223, opened 2025-04-08
Ak/read target data/221

### LINKED ISSUE #221: Create function to read in single target data files.
This function will act as the equivalent of `read_model_out_file()` and enable reading and validation of individual files.

This will allow us to validate only files that have been modified/added instead of the entire target dataset.

annakrystalli:
This PR resolves #221 

It enables us to read individual target data files which in turn will allow us to validate individula added or modified files instead of all target files.

This functionality also required the introduction of functionality for parsing hive-partitioned data from individual file paths as this was not avilable in arrow at the R level.

It also fixes a new error that seemed to have popped up with `match.arg` in `abs_file_path` when used in `check_file_exists()`.

~~PS I've requested a review from co-pilot to see what it comes back with!~~
Oh look like R is not supported yet...

### DIFF
--- a/man/extract_hive_partitions.Rd
+++ b/man/extract_hive_partitions.Rd
@@ -4,7 +4,7 @@
 \alias{extract_hive_partitions}
 \title{Extract Hive-style partition key-value pairs from a path}
 \usage{
-extract_hive_partitions(path, strict = FALSE)
+extract_hive_partitions(path, strict = TRUE)
 }
 \arguments{
 \item{path}{A character string of length 1: the path to a file or directory.}
--- a/man/is_hive_partitioned_path.Rd
+++ b/man/is_hive_partitioned_path.Rd
@@ -4,7 +4,7 @@
 \alias{is_hive_partitioned_path}
 \title{Check whether a path contains Hive-style partitioning}
 \usage{
-is_hive_partitioned_path(path, strict = FALSE)
+is_hive_partitioned_path(path, strict = TRUE)
 }
 \arguments{
 \item{path}{Character string. Path to a
--- a/man/read_target_file.Rd
+++ b/man/read_target_file.Rd
@@ -53,6 +53,7 @@ a tibble of contents of the target data file.
 Read a single target data file
 }
 \examples{
+\dontshow{if (requireNamespace("curl", quietly = TRUE) && curl::has_internet()) (if (getRversion() >= "3.4") withAutoprint else force)(\{ # examplesIf}
 # download example hub
 hub_path <- withr::local_tempdir()
 example_hub <- "https://github.com/hubverse-org/example-complex-forecast-hub.git"
@@ -63,4 +64,5 @@ read_target_file("time-series.csv", hub_path, coerce_types = "chr")
 # read in oracle-output file
 read_target_file("oracle-output.csv", hub_path)
 read_target_file("oracle-output.csv", hub_path, coerce_types = "chr")
+\dontshow{\}) # examplesIf}
 }


R/utils-hive-partitions.R
```diff
@@ -0,0 +1,175 @@
...
+  # Split the path
+  split_path <- fs::path_split(path)[[1]]
+
+  # Keep only the parts that contain an '=' sign — candidate key=value pairs
+  kv_pairs <- split_path[grepl("=", split_path)]
+
+  # Extract key-value pairs safely
+  purrr::map(kv_pairs, function(pair) {
+    # Split on the first '='
+    parts <- utils::strcapture("^([^=]+)=(.*)$",
```
zkamvar, line 91: I'm always so happy whenever I see this function being used in the wild 🤩 It's so useful and yet not many know about it

tests/testthat/test-read_target_file.R
```diff
@@ -0,0 +1,303 @@
...
+    ~ {
+      target <- gsub(" ", "_", .y)
+      path <- file.path(dir, paste0("target-", target, ".csv"))
+      arrow::write_csv_arrow(.x, file = path)
+    }
+  )
+  fs::dir_ls(dir)
+
+  # read in one of the time-series file with hub schema
+  file <- read_target_file("time-series/target-wk_inc_flu_hosp.csv", dir_hub_path)
```
zkamvar, line 133: I like this because it provides an avenue for hubs that do not want to combine targets into a single time series file.

R/read_target_file.R
```diff
@@ -0,0 +1,116 @@
...
+
+  # Convert to tibble
+  tbl <- tibble::as_tibble(tbl)
+
+  # Extract partition values and bind them
+  partition_row <- extract_partition_df(
+    target_file_path, schema, strict = TRUE
+  )
+  if (!is.null(partition_row)) {
+    tbl <- dplyr::bind_cols(tbl, partition_row)
```
zkamvar, line 95: This confused me a bit because I was trying to figure out why we were binding cols with a row of data.

I now understand that it's because you will only ever get one row of data with this function, but I would prefer to name it as "partition_df" to avoid that confusion. 

```suggestion
  partition_df <- extract_partition_df(
    target_file_path, schema, strict = TRUE
  )
  if (!is.null(partition_df)) {
    tbl <- dplyr::bind_cols(tbl, partition_df)
```

tests/testthat/test-read_target_file.R
```diff
@@ -0,0 +1,303 @@
...
+  dir <- fs::path(dir_hub_path, "target-data", "time-series")
+  fs::dir_create(dir)
+  split(dat, dat$target) |> purrr::iwalk(
+    ~ {
+      target <- gsub(" ", "_", .y)
+      path <- file.path(dir, paste0("target-", target, ".csv"))
+      arrow::write_csv_arrow(.x, file = path)
+    }
+  )
+  fs::dir_ls(dir)
```
zkamvar, line 130:
```suggestion
```

zkamvar (commented):
This looks great! I think you came up with a really clever solution to extract the data from the hive partitioned files and I really like to see all the tests you've added. 

One of my questions is: what is the use-case for setting `strict = FALSE` for `is_hive_partitioned_path()`? Considering that we set `strict = TRUE` for `read_target_file()`, it seems like that should be the default. When I ran `is_hive_partitioned_path("data/=US/year=2024/")`, I was a bit surprised to see that it returned `TRUE` even though it was malformed.

R/utils-hive-partitions.R
annakrystalli: Hehe! The first time I saw it was in one of your PRs a while back! I like it a lot now too!

annakrystalli:
Thanks for the review @zkamvar !

I've committed all your suggestions!

Re:
> One of my questions is: what is the use-case for setting strict = FALSE for is_hive_partitioned_path()? Considering that we set strict = TRUE for read_target_file(), it seems like that should be the default. When I ran is_hive_partitioned_path("data/=US/year=2024/"), I was a bit surprised to see that it returned TRUE even though it was malformed.

Fair point. I set the default value of `strict` to `TRUE` throughout 👍

zkamvar (approved):
Thank you for the hard work!

Non-blocking, but I would suggest to remove the `strict` argument in the future so that valid hives return TRUE, non-hives return FALSE, and invalid hives return an error. This will reduce the number of scenarios we have to test for and make it easier to maintain.

annakrystalli:
> Non-blocking, but I would suggest to remove the strict argument in the future so that valid hives return TRUE, non-hives return FALSE, and invalid hives return an error.

I can envisage a situation where someone might want to run a non strict check to determine more generally if the expectation for the hub are hive-partitioned files even if an individual file might be malformed (which will be addressed in validation). So for now at least I'd prefer to keep it. If we find that indeed there no use for this I'd be happy to remove it then.

PR #224, opened 2025-04-15
Validate hive-partition target file paths can be parsed correctly

### LINKED ISSUE #222: Create check function for checking hive partition file paths
When validating individual hive partitioned files, we need to ensure partitions can be correctly parsed before trying to read any files.

This should throw an error that stops execution if hive partitioning is malformed and should just skip if there is no partitioning detected at all.

annakrystalli:
This PR adds the `check_target_file_name()` function which ensures hive-partitioned target file paths can be parsed correctly. If no hive-partitiones are detected the check is skipped (technically it is not skipped as the overall check is implemented via a single call to `is_hive_partitioned_path()` but no result is necessary for non-hive partitioned paths so we return a `<check_info>` class object and report the check is skipped).

The PR resolves #222


annakrystalli:
Thanks for the speedy review @zkamvar !

> There's a misspelling in "partitions" in the tests, but they don't affect anything and are not user-facing, so this is good to go!

Doh! Fixed it!

PR #228, opened 2025-04-16
Add target dataset checks for shared file extensions

### LINKED ISSUE #227: Create check function that ensures files in multi-file target datasets share the same unique extension
Check that If files are being submitted in a multi-file dataset, the files in directory share the same unique extension.

This will be a directory level check that will happen once per target data type being submitted, not repeated for each individual target file being validated (if more than one is being submitted).

annakrystalli:
This PR resolves #227 

I ended up breaking up the check up into two individual checks:
1. Check that only a single dataset exists for a given target type (e.g. we are not mixing a single `time-series.csv` with a multifile `time-series` directory in the same hub).
2. Check that all target dataset file(s) share the same unique extension.

These checks should not be run on individual files (which a given PR may well contain multiples of) but **once** at the target type level before running further checks on individual data files. See https://github.com/hubverse-org/hubValidations/issues/229

NOTE: This PR requires the merging of https://github.com/hubverse-org/hubData/pull/85 before checks will pass


NEWS.md
```diff
@@ -6,6 +6,8 @@
 * Added `read_target_file()` function for reading in individual target data files.
 * Added target data validation checks:
   - `check_target_file_name()`: that a hive-partitioned target data file name can be correctly parsed.
+  - `check_target_dataset_unique()`: that a single unique target dataset exists for a given target type.
+  - `check_target_dataset_file_ext_unique()`: that file(s) in a target dataset share a single unique file extension.
```
LucieContamin, line 10: I will suggest to  add the information that only recognize target data file(s) are tested.
For example, 
```suggestion
  - `check_target_dataset_file_ext_unique()`: that target file(s) (`time-series`, `oracle-output`) in a target dataset share a single unique file extension.
```

R/check_target_dataset_file_ext_unique.R
```diff
@@ -0,0 +1,52 @@
+#' Check that all files of a given target type share a single unique file format
+#'
```
LucieContamin, line 2: Same as in NEWS, I think we can add the detail that only target data are concerned by the test. (It was not obvious to me and I tested with a badly name file with another extension and as expected got no error)

R/check_target_dataset_unique.R
```diff
@@ -0,0 +1,47 @@
...
+#' @export
+check_target_dataset_unique <- function(hub_path,
+                                        target_type = c(
+                                          "time-series",
+                                          "oracle-output"
+                                        )) {
+  target_type <- rlang::arg_match(target_type)
+  file_path <- file.path("target-data", target_type)
+
+  target_path <- hubData::get_target_path(hub_path, target_type)
```
LucieContamin, line 14: I did not get an error if I have both `time-series.csv` and a partitioned version of the file as `target=inc%20death/time-series0.parquet` + `target=inc%20hosp/time-series0.parquet`.

It seems that target_path is an empty character string for the partitioned version.

R/check_target_dataset_file_ext_unique.R
LucieContamin: And it does not seems to work if you have both partitioned and non-partitioned files. 
In fact, partitioned files do not seem to be identify:
For example, on my test I have:
```r
> hubValidations::check_target_dataset_unique(".")
<message/check_info>
Message:
No time-series target type files detected in target-data directory. Check skipped.
```

and my target-data folder is organized as:
```
|__ target-data
        |_____ target=inc%20death
                              |_______part-0.parquet (same if file called time-series0.parquet)
        |_____ target=inc%20hosp
                              |_______part-0.parquet (same if file called time-series0.parquet)
```

LucieContamin (commented):
Thanks for all the work.
I don't know if I misunderstood something or did something wrong, but I am having trouble having the validation working on partitioned files, is is expected? If so, I will remove some of my comments

R/check_target_dataset_file_ext_unique.R
zkamvar: @LucieContamin, the reason the partitioned files are not identified is because they need the partitions need to be in a folder called `time-series` (see [2025-01-06 RFC Target Data Formats](https://github.com/reichlab/decisions/blob/main/decisions/2025-01-06-rfc-target-data-formats.md#target-data-location) for a description of target data formats)

```
target-data
|__ time-series
        |_____ target=inc%20death
                              |_______part-0.parquet (same if file called time-series0.parquet)
        |_____ target=inc%20hosp
                              |_______part-0.parquet (same if file called time-series0.parquet)
```

LucieContamin: It's all good now, so please ignore my second comment, I will just suggests to update the documentation to make it clear what is tested

LucieContamin (approved):
Sorry for the previous issue, it was on my side, and it's all good now. I will just suggests to add more details in the documentation, but please feel free to ignore

R/check_target_dataset_file_ext_unique.R
LucieContamin: it's ok, I opened an issue in the hubdoc repo, let me know if I can help. I am happy to review or any other help

annakrystalli:
> My only other comment is not a blocker and more of a question (it's like the _opposite_ of an academic conference): I am _pretty sure_ that `hubData::get_target_path()` will return a single path for hive-partitioned data, is this the case?

Yes! For example if we're detailing with the `time-series` data, it will return `target-data/time-series`

R/check_target_dataset_file_ext_unique.R
annakrystalli: Yeah sorry about that @LucieContamin . The target data work is being rolled out on multiple fronts and the hubDocs documentation has not been updated with respect to some of the later decisions. Definitely an important high impact missing piece though! Perhaps we should discuss at the meeting whether to bump that up the priority scale.

NEWS.md
annakrystalli: I've added a comment in the NEWS. In the function documentation, the two target types are listed in the `target_type` argument options.

annakrystalli:
Thanks both @zkamvar and @LucieContamin  for the reviews!

PR #232, opened 2025-04-24
Optimise required values validation

annakrystalli:
This PR attempts to further optimised the validation of required values by implementing the following:
1. Use dplyr's `groub_by()` and associated `groub_indices()` and `group_rows()` to replace identifying rows with shared optional values using concatenation of their values via the slow `conc_rows()` function.
2. Kept masks of required values `req` as matrices.
3. Use `NA`s to mask non required values in `tbl`. 
4. Instead of splitting the `tbl` and `req` objects using `split()`, subsets of each the initial dataset are passed for processing to `missing_required()`.

I also added some additional tests that clarify expectations.

Interested to hear whether this improves @LucieContamin 's validations although in truth, the significant bottleneck will likely shift onto the joins part of the workflow.


LucieContamin (approved):
Thank you very much for the update. It is faster than previously on my tests. I profiled one of my test, and it seems that the new get_group_row and missing_required are now the major bottleneck. But we might need to double check

zkamvar (approved):
Looks great! If Lucie's happy with the speedup and the tests pass then it's all good. I especially appreciate the explanatory comments and the descriptive function names. 

I have two suggestions for minor speedups to replace `apply()` functions with `colSums()`, but these are absolutely not blockers. 


EDIT: I was curious about how this scales to hub-sized data (millions of rows) and found that it actually shaves off both time and memory:

``` r
mask <- data.frame(lapply(letters[1:12], \(x) runif(1e7) > .01))
one <- function(x) apply(x, 2, all)
two <- function(x, n) colSums(x) == n
tic <- Sys.time()
print(bench::mark(one(mask), two(mask, n = nrow(mask)), iterations = 30))
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 13
#>   expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
#>   <bch:expr>    <bch> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
#> 1 one(mask)     548ms  577ms      1.73    1.79GB     6.36    30   110     17.29s
#> 2 two(mask, n … 193ms  201ms      5.01  457.76MB     2.51    30    15      5.99s
#> # ℹ 4 more variables: result <list>, memory <list>, time <list>, gc <list>
Sys.time() - tic
#> Time difference of 24.81603 secs
```

<sup>Created on 2025-04-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

annakrystalli:
Thank you both!!

@LucieContamin , thanks for testing. I guess whether that's the bottleneck depends on the number of different optional value combinations that need to be tested for required value which is both a function of how compex the config is and how many optional value combinations the`tbl` contains so I guess it can still be the bottleneck in situations where the above complexity is high. 

In large but not as complex files I tested the bottleneck seemed to be the join to allocate model output data to model tasks (which is also why the function can be doing sth before completely ignoring a non required model task (as you reported in #203. All useful info though!

@zkamvar great suggestion on the use of `colSums`! Commited!

PR #233, opened 2025-04-30
Add check target colnames function

### LINKED ISSUE #215: Create check function to check target data column names
The check function(s) should check that:

## timeseries target data

time-series data should have the following columns:
 - an `observation` column
 - an optional `as_of` column
 - the other columns are valid task ID columns

 ## oracle output data

oracle-output data should have the following columns:
- an `oracle_value` column
- valid `task_id` columns

The task ID columns for now will be determined across all rounds

annakrystalli:
This PR resolves #215 (see issue for details) by introducing function `check_target_tbl_colnames()`.

The checks validates the columns in a target data file match expectations according to target type, as defined by our target data standards.

Note that this check does not ensure all columns that exist in other data files when dealing with multi-file target data exist in the file being checked, just that they are valid according to our standards. I'll open another issue for the checking columns in multi-file datasets for consistency.

The PR also adds some utilities for extracting target metadata useful for target data validation.

To be able to access example data throughout tests I also introduced a setup script that clones the example complex forecasting hub. This can also be sourced locally and will only re-run the cloning code if the temp example hub has been cleared. Makes developing tests locally much more efficient while still cleaning up during testsing.

I also made a small tweak the the function for generating check outputs so that `details` are collapsed into a single string so as to avoid repeating the overall check message if `details` has a length greater than 1L.


R/utils-target-metadata.R
```diff
@@ -0,0 +1,65 @@
...
+#'
+#' @param config_tasks a list representation of the `tasks.json` config file.
+#'
+#' @return A character vector of unique target task ID names. Post v5.0.0
+#' this should be a single task ID name.
+#' @export
+get_target_task_id <- function(config_tasks) {
+  get_target_metadata(config_tasks) |>
+    purrr::map_chr(~ names(.x[["target_keys"]])) |>
+    unique()
```
zkamvar, line 15: This allows hubs with implicit targets to be included.

```suggestion
  safe_names <- function(x) names(x) %||% ""
  res <- get_target_metadata(config_tasks) |>
    purrr::map_chr(~ safe_names(.x[["target_keys"]])) |>
    unique()
  if (identical(res, "")) NULL else res
```

zkamvar (changes requested):
Thank you! This works as expected and when I tested this with the flusight hub data, I was able to get a passing result 🎉 

I'm also 100% on board with the test fixture! This definitely makes things much easier to work with.  

## Changes requested

I tried `get_target_task_id()` with the variant nowcast hub, but got an error because this hub has implicit targets. I made a suggestion that will fix this problem. 

Additionally, there is one thing missing in this check. We need to verify that the data have the minimum number of required columns:

> Each row of the data set is a unit of observation and the columns consist of:
>
> - task ID variables that uniquely define the unit of observation. This must include at least one column representing the date.
> - an observation column that records the observed value
>
> [docs.hubverse.io: Time Series Data](https://docs.hubverse.io/en/latest/user-guide/target-data.html#time-series)

The current validations would declare time series data with only `target` and `observation` columns to be valid. 

A valid time series data set should have a minimum of 3 columns if there are explicit targets and 2 columns if there are implicit targets.

LucieContamin (approved):
thank you very much! It's all look good to me! I did not test Zhian update but I trust it. 
For the "date" requirement, I am not sure how you want to implement it.

R/utils-target-metadata.R
annakrystalli: Thanks for catching this very obvious omision! I've added a very slightly modified version for this and also a bunch of tests too.

annakrystalli:
Thank you both for you comments! Some excellent catches @zkamvar !

I've incorporate both your comments by:
1. Adding your suggested change as well as explicit tests for checking this.
2. Add a minimum number of columns for time-series data

As for the date column, fear not! This will be tackled in a check that validates column data times...coming soon to a PR near you! 😆

annakrystalli:
Oh man! It looks like the lintr package has changed default behaviour again and the failure in unrelated to the changes.

Do you mind if I open an issue to tackle it in a separate PR?

zkamvar (approved):
Looks good to me!

PR #242, opened 2025-07-11
Add `check_target_tbl_output_type_ids()` fn / 219

### LINKED ISSUE #219: Create check function to validate target oracle-output data `output_type_id` values #217
oracle-output data `output_type_id` values should meet the following criteria
- for output types `"mean"`, `"median"`, `"quantile"`,  and `"sample"`, `output_type_id`s should be `NA`.
- for output types `"cdf"` and `"pmf"`:
    - `output_type_id`s should match the `output_type_id`s defined in the config. 
    - Additionally, each unique combination of task ID values should have a row for each expected `output_type_id` defined in the config.

annakrystalli:
This PR resolves #219 by adding a function that checks that output type id values in oracle-output target data (if present) matches expected values.

- For any summary statistic/sample output types (i.e. quantile, mean, median and sample), output type IDs must have `NA` values.
- For any distributional output types (i.e. cdf or pmf),  observations (as defined by the observable unit) must contain all output_type IDs

I also made some minor renaming in some testing helper fns (change any reference to an "inferred" target (i.e. a target in a hub that only has a single target, has no target task ID and column and the target is defined in the config only) to "global".


LucieContamin (approved):
Thank you! It looks good, I just review the code and run the test without issue. 

I just have a question on how and why we define "Global target", but I don't think it's blocking this PR.

PR #244, opened 2025-07-21
Add check_target_tbl_coltypes fn / 216

### LINKED ISSUE #216: Create check function to validate target data column types
The check function(s) should check that:

## timeseries target data

time-series data should have the following data types:
 - a numeric `observation` column
 - an optional `as_of` date column
 - the other columns  should match the respective task ID column data types
 - At least one column excluding `as_of` should be able to be parsed as date. `date_col` argument might be required if column is from a partition and does not match any task ID column names which could be used to determine data type.

 ## oracle output data

oracle-output data should have the following columns:
- a numeric `oracle_value` column
- the other columns  should match the respective task ID column data types

annakrystalli:
This PR resolves #216 by adding a check that ensures target data column data types match the expected schema for the target type.

Note that eventually the schema will be determined from the target-data.json config but that doesn't affect the functionality here as it will be incorporated in `hubData::create_timeseries_schema` and `hubData::create_oracle_output_schema()` functionality.

It also includes a number of minor changes made to fix new notes/lintr issues that popped up:
- Fix minimum version of R to use base pipe
- Fix some nolint comments to work with air formatting
- Add namespacing to lubridate period class and re-export `[hubUtils::get_hub_derived_task_ids](https://github.com/hubverse-org/hubValidations/pull/244/commits/ebfc2bbaf2651ed466cfb5f2c9ced9792eb1483c)` to fix roxygen complaints


R/read_target_file.R
```diff
@@ -27,10 +27,13 @@
...
-                             coerce_types = c("hub", "chr", "none"),
-                             date_col = NULL,
-                             na = c("NA", "")) {
+read_target_file <- function(
+  target_file_path,
+  hub_path,
+  coerce_types = c("target", "chr", "none"),
+  date_col = NULL,
+  na = c("NA", "")
+) {
```
LucieContamin, line 36: The documentation needs to be update for the `coerce_types`, it's still "hub", "chr" or "none" in the documentation.

LucieContamin (changes requested):
Thank you very much, it looks good, I think it's just missing a small documentation update. (please ignore if I misunderstood)

R/read_target_file.R
annakrystalli: Excellent catch! Fixed in dee3fd3240f365d8257a576628aa87fffbd4d8d9

LucieContamin (approved):

PR #251, opened 2025-07-22
Add `check_target_tbl_rows_unique()` fn / 247

### LINKED ISSUE #247: Create check_target_tbl_rows_unique function
The function should check that there are no duplicate rows in target data files being validated, equivalent to `check_tbl_rows_unique()`

If datasets are versioned, in `time-series` data, multiple observations are allowed so long as they have different `as_of` values so the `as_of` column should be included when determining duplicates. In `oracle-output` data though, there should be only a single observation regardless of the `as_of` value so the column should not be included when determining duplicates.

This function should:
1. Be a subfunction of `validate_target_tbl()`

annakrystalli:
This PR resolves #247 by introducing a function that checks target files for duplicates.

Note that if target datasets are versioned, in `time-series` data, multiple observations are allowed so long as they have different `as_of` values so the `as_of` column is included when determining duplicates. 
In `oracle-output` data though, there should be only a single observation regardless of the `as_of` value so the column is not included when determining duplicates.


LucieContamin (approved):
thanks! It looks good to me!

PR #252, opened 2025-07-29
Add check_target_dataset_rows_unique fn /246

### LINKED ISSUE #246: Create check_target_dataset_unique_obs function
While individual target data file validation checks for duplicate observations, in multifile datasets this does not guarantee that duplicates do not exist.

It would be good to have a function that checks that observations across each entire dataset is unique as part of the overall target dataset health checks.

If datasets are versioned, in `time-series` data, multiple observations are allowed so long as they have different `as_of` values so the `as_of` column should be included when determining duplicates. In `oracle-output` data though, there should be only a single observation regardless of the `as_of` value so the column should not be included when determining duplicates.

This function should:
1. Be a subfunction of `validate_target_dataset()`
2. Because it could be expensive, should be able to be skipped during incoming file validation but could instead be deployed as a separate periodic health check via a separate workflow.

annakrystalli:
This PR resolves #246 by adding a function that checks that each row in a target dataset is unique. The function is designed to be used as part of overall target data integrity check, ensuring there are no duplicates over the entire dataset, not just a single file being validated.

Note that if target datasets are versioned, in `time-series` data, multiple observations are allowed so long as they have different `as_of` values so the `as_of` column is included when determining duplicates. 
In `oracle-output` data though, there should be only a single observation regardless of the `as_of` value so the column is not included when determining duplicates.

The PR also:
- adds some testing functions for standing up hubs with hive partitioned target datasets
- Fixes a tidyselect warning that started popping up in the tests
- contains some minor air reformatting of files that were saved.


R/check_target_tbl_oracle_value.R
```diff
@@ -226,7 +226,7 @@ check_oracle_value_cdf_crossing <- function(
       0L
   ) |>
     filter(.data[["invalid"]]) |>
-    select(-.data[["invalid"]]) |>
+    select(-all_of("invalid")) |>
```
annakrystalli, line 229: This fixes tidyselect complaining

PR #254, opened 2025-07-30
Add `check_target_file_read()` fn / 248

### LINKED ISSUE #248: Create `check_target_file_read()` function
This function should check that the target file being validated can be read successfully.

For csv files which do not contain a schema like parquet and sometimes need to be coerced to ensure correct data types, the coercion to target schema also represents a test of whether the file can be read with the appropriate schema.

annakrystalli:
This PR resolves #248 by adding a functions that checks that the target files being validated can be successfully read.

I also made a small modification to  `check_file_read()`, removing an unnecessary assignment.


LucieContamin (approved):
Thank you! It looks good! 

I just have a remark that does not stop the merge, but I wonder if we can create a function for the `check_X_read()` functions to avoid redundancy. I could not think of a solution.

annakrystalli:
Thanks! I thought of that but the use different functions and return different messages so we would end up with a much longer function full of if statements and more complicated documentation too. So I feel it's cleaner this way. It's a pretty short function too

annakrystalli:
Also need to look into the tests failing on windows 😒

Issue #225, opened 2025-04-15
Add all target data checks to check_table.csv


annakrystalli:
Add all target data checks to `check_table.csv` and print in higher level target data validation function documentation

PR #255, opened 2025-08-07
Add validate_target_file / 250

### LINKED ISSUE #250: Create validate_target_file function
This function will mirror the checks performed by`validate_model_file()` for model outputs and perform all the checks related to target file properties being validated. i.e. all `check_target_file_*()` check functions

annakrystalli:
‼️ Depends on https://github.com/hubverse-org/hubUtils/pull/238 being merged first for tests to pass!

This PR resolves #250 by adding function `validate_target_file()` for validating file level properties of a target data file.

The PR also begins to address #225 by adding the information related to the checks deployed by `validate_target_file()` to `inst/check_table.csv` and presenting the info in a table in the `validate_target_file` man page.

Note that  the roxygen documentation in `validate_pr()` and `validate_submission()` was also modified to ensure the correct checks where being shown in each fns info table. That's the only part of those files that need reviewing, the rest is formatting.


R/validate_submission.R
```diff
@@ -30,7 +30,9 @@
 #'
 #' ```{r, echo = FALSE}
 #' arrow::read_csv_arrow(system.file("check_table.csv", package = "hubValidations")) %>%
-#' dplyr::filter(.data$`parent fun` != "validate_model_metadata", !.data$optional) %>%
+#' dplyr::filter(
+#' !.data$`parent fun` %in% c("validate_model_metadata", "validate_target_file"),
+#' !.data$optional) %>%
 #'   dplyr::select(-"parent fun", -"check fun", -"optional") %>%
```
annakrystalli, line 36: Only the above block needs reviewing in this file. The rest of the changes are formatting.

R/validate_target_file.R
```diff
@@ -0,0 +1,95 @@
...
+  checks$target_file_ext <- try_check(
+    check_target_file_ext_valid(
+      file_path = file_path
+    ),
+    file_path
+  )
+  if (is_any_error(checks$target_file_ext)) {
+    return(checks)
+  }
+
```
LucieContamin, line 88: What don't we add `check_target_dataset_file_ext_unique()` in this function, too?

LucieContamin (changes requested):
Thank you for the update, I mostly have questions about the new function. Also, should failing tests should also be added in the tests?

R/validate_target_file.R
```diff
@@ -0,0 +1,95 @@
+#' Validate file level properties of a target data file.
+#'
+#' @param file_path A character string representing the path to the target data
+#'  file relative to the `target-data` directory.
+#' @param validations_cfg_path Path to YAML file configuring custom validation checks.
+#' If `NULL` defaults to standard `hub-config/validations.yml` path. For more details
+#' see [article on custom validation checks](
+#' https://hubverse-org.github.io/hubValidations/articles/deploying-custom-functions.html).
```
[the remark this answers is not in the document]
annakrystalli: It is deployed exactly the same as model output validation functions as described here https://hubverse-org.github.io/hubValidations/articles/deploying-custom-functions.html#default-configuration, i.e.:

- **`<name-of-caller-function>`:** One of `validate_model_data`, `validate_model_metadata` and `validate_model_file` depending on the function the custom check is to be included in.
    - **`<name-of-check>`:** The name of the check. This is the name of the element containing the result of the check when `hub_validations` is returned (required).
        - **`fn`:** The name of the check function to be run, as character string (required).
        - **`pkg`:** The name of the package namespace from which to get check function. Must be supplied if function is distributed as part of a package.
        - **`source:`** Path to `.R` script containing function code to be sourced. If relative, should be relative to the hub's directory root. Must be supplied if function is not part of a package and only exists as a script.
        - **`args`:** A yaml dictionary of key/value pairs of arguments and their values to be passed to the custom function. Values can be yaml lists or even executable R code (optional).
    

I've added this issue to update available objects within target data validation functions when they are added: https://github.com/hubverse-org/hubValidations/issues/258

inst/check_table.csv
```diff
@@ -33,6 +33,9 @@ metadata_file_ext,The metadata file has correct extension (yaml or yml).,TRUE,ch
 metadata_file_location,The metadata file has been saved to correct location.,TRUE,check_failure,validate_model_metadata,check_metadata_file_location,,FALSE
 metadata_matches_schema,The contents of the metadata file match the hub's model metadata schema,TRUE,check_error,validate_model_metadata,check_metadata_matches_schema,,FALSE
 metadata_file_name,The metadata filename matches the model ID specified in the contents of the file.,TRUE,check_error,validate_model_metadata,check_metadata_file_name,,FALSE
+target_file_exists,File exists at `file_path` provided,TRUE,check_error,validate_target_file,check_file_exists,,FALSE
+target_partition_file_name,Hive-style partition file path segments are valid and can be parsed successfully. Skipped if target dataset not hive-partitioned.,TRUE,check_error,validate_target_file,check_target_file_name,,FALSE
+target_file_ext,Target data file extension is valid.,TRUE,check_error,validate_target_file,check_target_file_ext_valid,,FALSE
```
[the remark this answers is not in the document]
annakrystalli: Fixed in a0c8040d93110a12b7f69603934f89c1ccfc0db2

annakrystalli:

Hopefully I've addressed all your questions. We just need to get https://github.com/hubverse-org/hubUtils/pull/238 merged in so we can deploy the tests properly

R/validate_target_file.R
LucieContamin: I thought `check_target_dataset_file_ext_unique()` was extracting the file extension of each file and checking it was in the expected format (for example one unique csv file). If I understand it, we don't open/read the files, so for me it was a file level function. Did I misunderstood file level and dataset wide ? (sorry if so)

LucieContamin (commented):
Thank you, it all looks good, I just have a following question about `check_target_dataset_file_ext_unique()` . But, if you need to merge the PR, we can go back to it later on.

annakrystalli:
No it's checking that in multifile target datasets, all files share a single unique extension. That's why it's a dataset property

LucieContamin (approved):
Thanks for the quick answer!

PR #261, opened 2025-08-27
Add validate_target_data function / 249

### LINKED ISSUE #249: Create validate_target_data function
This function will mirror the checks performed by`validate_model_data()` for model outputs and perform all the checks related to the content of target files being validated. i.e. all `check_target_tbl_*()` check functions

annakrystalli:
This PR resolves #249 by adding the `validate_target_data()`  to perform validation checks on the contents of target data files. 

More specifically:

- Introduces `validate_target_data()` as the top-level validation function.
- Wraps and orchestrates content validation checks on individual target files.
- Updates `check_table.csv` with the data level checks.
- Additionally adds a useful helper functions in `tests/testthat/helper-early-returns.R` for testing the expected early return behaviour of  `validate_target_data()`.


R/validate_target_data.R
```diff
@@ -0,0 +1,186 @@
...
+      na = na,
+      output_type_id_datatype = output_type_id_datatype,
+      file_path = file_path,
+      hub_path = hub_path
+    ),
+    file_path
+  )
+  if (is_any_error(checks$target_tbl_coltypes)) {
+    return(checks)
+  }
```
LucieContamin, line 118: Can I ask why this one stop the checking and returns an error? Will it cause issue in the next checks if it fails?

R/validate_target_data.R
```diff
@@ -0,0 +1,186 @@
...
+      target_tbl_chr = target_tbl_chr,
+      target_type = target_type,
+      file_path = file_path,
+      hub_path = hub_path
+    ),
+    file_path
+  )
+  if (is_any_error(checks$target_tbl_values)) {
+    return(checks)
+  }
```
LucieContamin, line 155: Same as my other question, Can I ask why this one stop the checking and returns an error? Will it cause issue in the next checks if it fails?

LucieContamin (approved):
Looks good to me! thank you!! I just have 2 quick questions, feel free to ignore and we can talk about it later on if necessary

R/validate_target_data.R
annakrystalli: This one affects the output type id check. The task ID values need to be correct because the output type IDs are checked for validity and completeness by grouping by task IDs. So if task IDs contain invalid values, we can't guarantee the output type id check result is valid.

R/validate_target_data.R
annakrystalli: The oracle value check assumes that the column is the correct data type so is affected. The ts_target check also assumes the target column is the correct data type. But more importantly column data type is an important property that custom checks would also assume has already been checked and is correct.

annakrystalli:
Thanks for the speedy review @LucieContamin !

LucieContamin:
Thanks for the additional information, I was not totally sure of all the impacts. It's all clear now!

PR #266, opened 2025-08-28
Add `<target_validations>` subclass and custom print method /265

### LINKED ISSUE #265: `file_path` in checks of multifile target datasets should be recorder as relative to `target-data` dir, not the basename
Currently the path recorded in `file_path` (and retuned in the `where` property of the output (`hub_check` object) of each check performed in [`validate_target_file()`](https://hubverse-org.github.io/hubValidations/dev/reference/validate_target_file.html) and [`validate_target_data()`](https://hubverse-org.github.io/hubValidations/dev/reference/validate_target_data.html) is the basename of the target file path. This is fine for single file target data sets but is awkward for multi-file datasets.

In hive-partitioned dataset where multiple files might be called `part-0.parquet` the name is uninformative and vague and would actually cause problems when printing a `hub_validations` object that contains the results of validations for multiple files.

This is because the print method groups the results of individual checks by the path in `where`. As such, if all files are called `part-0.parquet` all checks will be printed under the single filename.

To fix this issue, the `file_path` recorded, for now atleast, should be relative to the `target-data` directory.

NOTE: I feel the current way we store and print multi-file validation results is not ideal but I will present a more detailed proposal on how i feel we can improve it once the target data validation work is done.

annakrystalli:
## Background

Currently the path used when printing `hub_validations` objects is the basename of the relative path stored in the `where` property of each check. 

For `model-output` files (which include the round and model IDs in the filename) and for single file target data sets this is fine, but is awkward for multi-file datasets (e.g. see outputs of [`validate_target_file()`](https://hubverse-org.github.io/hubValidations/dev/reference/validate_target_file.html) and [`validate_target_data()`](https://hubverse-org.github.io/hubValidations/dev/reference/validate_target_data.html).

In hive-partitioned dataset in particular where multiple files might be called `part-0.parquet`, the printed file name is not only uninformative and vague but could actually cause problems when printing a `hub_validations` object that contains the results of validations for multiple files.

This is because the print method groups the results of individual checks by the path **after** it's been reduced to it's basename. As such, if all files are called `part-0.parquet` all checks will be printed under the single filename, even if they apply to multiple files.

## Fix

This PR resolves #265 (described above) by adding a new `target_validations` subclass of `hub_validations` to enable a target validations specific print method. 

## Changes Made

- Added `target_validations` S3 subclass inheriting from `hub_validations`
- Added `new_target_validations()` and `as_target_validations()` constructor functions
- Added `combine.target_validations()` method for concatenating target validation objects
- Converted `get_filenames()` to S3 generic with `get_filenames.default()` and `get_filenames.target_validations()` methods which returns the full relative path instead of just the basenames. This enables `target_validations` objects to display full file paths in print output (vs basenames for `hub_validations`)
- Added tests for all new functionality
- Re-organised functions across files to be more intuitive

NOTE: I feel the current way we store and print multi-file validation results is not ideal but I will present a more detailed proposal on how I feel we can improve it once the target data validation work is done.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -51,11 +51,12 @@ reference:
 
 
 - title: "`<hub_validations>` methods"
-  desc: "Methods for `<hub_validations>` S3 objects."
+  desc: "Methods for `<hub_validations>` S3 objects and subclasses."
   contents:
   - starts_with("print")
   - starts_with("combine")
   - contains("hub_validations")
+  - contains("target_validations")
 
 navbar:
  components:


R/combine.R
```diff
@@ -0,0 +1,51 @@
+#' Concatenate `hub_validations` S3 class objects
```
annakrystalli, line 1: This first section is just moving the existing method to a new dedicated file

R/print.R
```diff
@@ -132,10 +81,15 @@ hub_validation_theme <- list(
     fmt = function(x) {
       cli::col_magenta(
         paste0(
```
annakrystalli, line 83: This is just reformatting

R/print.R
```diff
@@ -146,21 +100,54 @@ apply_cli_span_class <- function(x, class = "check_name") {
...
   } else {
     filenames
   }
 }
+
+#' @export
+#' @noRd
+#' @keywords internal
+get_filenames.target_validations <- function(x, unique = FALSE, ...) {
+  # Use full path instead of basename for target validations
```
annakrystalli, line 146: This line is the main difference!

R/utils-validation_classes.R
```diff
@@ -0,0 +1,24 @@
+validate_internal_class <- function(
```
annakrystalli, line 1: This was just moved to it's own file from a different file so no need for review. The only thing added was the `"target_validations"` as an option of class to validate elements of an object as

R/utils-validation_classes.R
```diff
@@ -0,0 +1,24 @@
...
+  x,
+  class = c(
+    "hub_check", # nolint
+    "hub_validations",
+    "target_validations"
```
annakrystalli, line 6: This is the only addition

tests/testthat/_snaps/validate_target_data.md
```diff
@@ -38,7 +38,7 @@
       res_ts
     Message
       
-      -- part-0.parquet ----
+      -- time-series/target=wk%20flu%20hosp%20rate/part-0.parquet ----
```
annakrystalli, line 41: Much better!! 🎉

PR #268, opened 2025-09-04
Ak/validate target pr/264

### LINKED ISSUE #264: Add `validate_target_pr()` fn
The function's purpose is to orchestrate the complete validation of all target data files which have been modified/added in an individual PR and is the equivalent to `validate_pr()` for model outputs. It is a function designed to be run exclusively against the contents of a pull request.

This would involve deploying the following functions:

- `validate_target_dataset()`: validating dataset level properties for any target types which have files that have been modified/added. Run once for each target type involved.
- `validate_target_submission()`: validating the individual files that have been modified/added. Run once for each file involved in the PR.

annakrystalli:
This PR resolves #264 by adding the `validate_target_pr()` function.

The function's purpose is to orchestrate the complete validation of all target data files which have been modified/added in an individual PR and is the equivalent to `validate_pr()` for model outputs. It is a function designed to be run exclusively against the contents of a pull request.

The function deployes the following functions:

- `validate_target_dataset()`: validating dataset level properties for any target types which have files that have been modified/added. Run once for each target type involved.
- `validate_target_submission()`: validating the individual files that have been modified/added. Run once for each file involved in the PR.

It also checks the hub's config files before performing any checks.

I also discovered a smarter way of checking different PRs by only cloning the ci test hub once and just switching between branches for testing different PRs so I plan to refactor the `validate_pr()` functions too to be more efficient. That can wait though.


R/validate_target_pr.R
```diff
@@ -0,0 +1,366 @@
...
+}
+
+is_target_type_file <- function(
+  path,
+  target_type = c("time-series", "oracle-output")
+) {
+  target_type <- rlang::arg_match(target_type)
+  is_target_type_single_file(path, target_type) ||
+    is_target_type_dir_file(path, target_type)
+}
```
LucieContamin, line 355: Not a blocking comment, but I am not sure I understand the logic here. Would it be possible to explain the logic, especially on the `is_target_type_single_file()`

LucieContamin (commented):
Thank you very much! It looks good! I just have some minor comments about the documentation of the function that might be a little confusing.

Issue #332, opened 2026-04-02
Bug in `validate_submission()` for samples output?


sbidari:
**Summary**

- validate_submission() fails on files with sample outputs and multiple model tasks (one per target for a round_id). 
- Error indicates validation steps expect one `mt_id` per group but files with two targets result in two unique `mt_id` values. 

**Error**
```
█ [spl_non_compound_tid]: EXEC ERROR: Error in dplyr::summarise(hash_tbl, n =
  dplyr::n_distinct(.data$hash_non_comp_tid), : ℹ In argument: `mt_id =
  unique(.data$mt_id)`. Caused by error: ! `mt_id` must be size 1, not 2. ℹ To
  return more or less than 1 row per group, use `reframe()`.
```

[Link](https://github.com/CDCgov/rsv-forecast-hub/actions/runs/23877786220/job/69624462350#step:8:174) to error on CI (contains other unrelated failures)

sbidari:
On further investigation, [this](https://github.com/hubverse-org/hubValidations/blob/ffde972b828a4cae13e44cd2ce6531d22af51f77/R/check_tbl_spl_non_compound_tid.R#L52) is the exact line in code that triggers the error for our setup. Switching `dplyr::summarise()` with `dplyr::reframe()` as suggested in the error message above allows the `check_tbl_spl_non_compound_tid` check to proceed and pass for our forecast file and hub configuration. 

I am not sure if this change is consistent with the intended validation logic or if it is circumventing an issue with our configs/data file. If this is the expected fix, I am happy to open a PR.

I’d also appreciate clarification on modeling-task semantics for multi-target hubs (like CovidHub, RSVHub). Do you think of targets as a part of compound task id sets or if targets define a separate model task, with its own compound task id sets. I suspect this distinction is central to the `mt_id` conflict behind this error.

annakrystalli:
Thanks for reporting this @sbidari!

## Root cause

The exec error is caused by sample `output_type_id`s spanning multiple model tasks in the submission. This should never happen. Different model tasks can have different sample configurations (`compound_taskid_set`, `min/max_samples_per_task`, etc.), so samples should be entirely independent across model tasks.

To answer your question about modelling task semantics for multi-target hubs: `target` is not treated specially during validation. It's just a task ID like any other. When targets define separate model tasks, each model task has its own sample configuration and samples should be contained within a single model task.

## RSV Hub misconfiguration

[...]

PR #336, opened 2026-04-06
Add `check_tbl_spl_mt_unique` to validate sample IDs don't span model tasks

### LINKED ISSUE #333: Add `check_tbl_spl_mt_unique` check to validate sample `output_type_id`s don't span model tasks
## Problem

Sample `output_type_id`s (sample indices) should be unique to a single model task within a round. Different model tasks can have different sample configurations (`compound_taskid_set`, `min/max_samples_per_task`, etc.), so samples should be entirely independent across model tasks.

Currently there is no check enforcing this. When sample IDs do span model tasks, it causes a downstream exec error in `check_tbl_spl_non_compound_tid` (see #332).

## Background

Investigating #332 revealed a deeper inconsistency in how `compound_taskid_set` is handled when task IDs (e.g. `target`) differentiate model tasks:

- When a task ID is single-valued within a model task (because it's what distinguishes the MTs), the detection logic in `check_tbl_spl_compound_taskid_set` **cannot** determine whether it's compound or just incidentally constant. It becomes invisible to validation.
- This means a correctly-structured submission (per-MT sample IDs, target effectively fixed) and a misstructured one (global sample IDs, target varying across MTs) can **both pass** validation, but for different reasons (one matches the config, the other is detected as "coarser" and allowed through).
- The config's intent for such task IDs is unenforceable at validation time within individual MTs.

Full enforcement requires coordination between:
- **Config validation** (hubAdmin): ensuring task IDs that differentiate MTs are in `compound_taskid_set`. See https://github.com/hubverse-org/hubAdmin/issues/157
- **Submission validation** (this check): ensuring samples don't cross MT boundaries
- **Coarser warning**: see #334

## Proposal

Add a new check `check_tbl_spl_mt_unique` that:

1. Runs **upstream** of the existing sample checks (before `check_tbl_spl_compound_taskid_set`), as an early-return error check.
2. Uses the `hash_tbl` from `spl_hash_tbl()` (or similar logic) to detect whether any `output_type_id` appears under multiple `mt_id`s.
3. Returns a clear validation error identifying the offending sample IDs and which model tasks they span.

Additionally, `check_tbl_spl_non_compound_tid` should explicitly handle this condition with an informative error message instead of the current cryptic `dplyr::summarise()` crash.

## Context

Discovered via [CDCgov/rsv-forecast-hub](https://github.com/CDCgov/rsv-forecast-hub) (#332) where a submission used sample IDs 1-200 globally across two model tasks (one per target).

annakrystalli:
## Summary

- Adds new `check_tbl_spl_mt_unique` check that validates individual sample `output_type_id`s do not span multiple model tasks. Runs upstream of existing sample checks with early return on error.
- Adds defensive error handling in `check_tbl_spl_non_compound_tid` to produce an informative error message instead of the cryptic `dplyr::summarise()` crash reported in #332.
- Adds test hub (`hub-spl-multi-mt`) with two model tasks both having sample output types.
- Increases cyclomatic complexity limit to 25 in `.lintr`.

## Test plan

- [x] New tests for `check_tbl_spl_mt_unique`: valid single-MT samples pass, samples spanning MTs error, skip when no v3 samples
- [x] New test for `check_tbl_spl_non_compound_tid` defensive error when samples span MTs
- [x] Existing sample check tests pass
- [x] Full test suite passes (1431 tests, 0 failures)
- [x] Lintr clean
- [x] Verified against CDCgov/rsv-forecast-hub submission that triggered #332

Closes #333


R/check_tbl_spl_mt_unique.R
```diff
@@ -0,0 +1,121 @@
+#' Check that individual sample output_type_ids do not span multiple model tasks
+#'
+#' @param tbl a tibble/data.frame of the contents of the file being validated.
+#' Column types must **all be character**.
+#' @inherit check_tbl_colnames params
+#' @inherit check_tbl_colnames return
+#' @inheritParams expand_model_out_grid
+#' @param derived_task_ids Character vector of derived task ID names (task IDs whose
+#' values depend on other task IDs) to ignore. Columns for such task ids will
+#' contain `NA`s. Defaults to extracting derived task IDs from hub `task.json`. See
```
LucieContamin: Not sure I understand the " Columns for such task ids will contain `NA`s." part here

LucieContamin (approved):
It all looks good, I made a small comment for more documentation but please feel free to ignore.

R/check_tbl_spl_mt_unique.R
annakrystalli: Good point! That sentence was exposing an implementation detail without explaining the user-facing effect. I've simplified it to just say "to ignore during validation" in 3113f43. Opened #339 to apply the same clarification to the other functions that use this parameter.

PR #337, opened 2026-04-06
Warn when detected `compound_taskid_set` is coarser than configured

### LINKED ISSUE #334: Consider warning when detected `compound_taskid_set` is coarser than configured
## Background

`check_tbl_spl_compound_taskid_set` currently allows submissions with a **coarser** `compound_taskid_set` than what's configured in `tasks.json`. This is statistically sound: a coarser dependence structure (more jointly varying dimensions) can always be sliced more finely for evaluation.

However, in practice, a coarser detected `compound_taskid_set` can also indicate a **misconfiguration** or a **submission error** rather than a deliberate modelling choice.

## Case study

In [CDCgov/rsv-forecast-hub](https://github.com/CDCgov/rsv-forecast-hub) (#332), the config specifies `compound_taskid_set: ["location"]` but the submission's detected compound_taskid_set is `character(0)` (nothing is fixed, one giant joint distribution across everything). This was not an intentional modelling choice but a submission error (sample IDs used globally instead of per compound group).

The "coarser is OK" policy let this through silently, which then caused a downstream exec error.

## Proposal

Add a **warning** to the output of `check_tbl_spl_compound_taskid_set` when the detected `compound_taskid_set` is coarser than configured. The check would still **pass**, but the warning would alert hub administrators to investigate whether the coarser structure is intentional.

This is especially important for **single model task hubs**, where the proposed `check_tbl_spl_mt_unique` check (#333) would not catch sample ID misuse (since there's only one model task to span).

## Related

- #332: original bug report
- #333: new `check_tbl_spl_mt_unique` check
- https://github.com/hubverse-org/hubAdmin/issues/157: config validation for `compound_taskid_set`

annakrystalli:
## Summary

- `check_tbl_spl_compound_taskid_set` now attaches a check-level warning when the detected `compound_taskid_set` is coarser than configured. The check still passes (coarser is statistically valid), but the warning alerts hub administrators to investigate whether the coarser structure is intentional or indicates a misconfiguration.
- Adds `detect_coarser_compound_taskid_set` helper in `compound_taskid-utils.R`.
- This is especially important for single model task hubs where `check_tbl_spl_mt_unique` (#333) cannot catch sample ID misuse.

Closes #334


R/check_tbl_spl_compound_taskid_set.R
```diff
@@ -76,6 +91,7 @@ check_tbl_spl_compound_taskid_set <- function(
     details = compile_msg(compound_taskid_set),
     errors = compile_errors(compound_taskid_set),
     error = TRUE,
+    warnings = coarser_warnings,
```
LucieContamin: Small question, will this warning cause the validation to "fail", meaning will it cause a validation PR to fail if this warning is trigger? If so, should we add it in the documentation? Personally, in SMH, we return that information too but as a note to avoid PR failure.  I am not against the warning, I just think if it might cause PR failure it should be good to document

R/compound_taskid-utils.R
```diff
@@ -261,3 +261,38 @@ get_comp_tid_sample_ids <- function(is_compound_taskid, output_type_ids) {
...
+      if (length(det) < length(conf) || !all(conf %in% det)) {
+        missing_comp_tids <- setdiff(conf, det) # nolint: object_usage_linter
+        det_label <- if (length(det) == 0L) "none" else det # nolint: object_usage_linter
+        msg <- cli::format_inline(
+          "Modeling task {mt_idx}: detected {.var compound_taskid_set} ",
+          "({.val {det_label}}) is coarser than configured ",
+          "({.val {conf}}). ",
+          "Configured compound task ID{?s} {.val {missing_comp_tids}} ",
+          "{?is/are} not fixed within samples. ",
+          "This may indicate a misconfiguration or submission error."
```
LucieContamin: Should we add something to say that the file is still valid before the "may indicate ...". It might be only me, but I feel like it should be clear that it may be an issue but it may also be all good, and more often it's probably all good.

R/check_tbl_spl_compound_taskid_set.R
LucieContamin: As the warning does not appear by default, should we update the message if the warning is not NULL to indicate that the compound task ID is coarser and more information are available in the warning?

LucieContamin (approved):
It all looks good, thank you! I just add a question and made some minor comments about documentation, please feel free to ignore.

R/check_tbl_spl_compound_taskid_set.R
annakrystalli: Good question. No, this won't cause PR failure. Validation warnings are informational only; `check_for_errors()` (which drives PR pass/fail) only aborts on `check_failure`, `check_error`, `check_exec_error` or `check_exec_warn` objects. You're right that this deserves documentation though, so I've added a note to the `capture_validation_warning()` docs.

R/compound_taskid-utils.R
annakrystalli: Thanks for flagging. I think the warning is a useful flag either way because it's important for hub admins to know that coarser samples exist, and also for modellers to be reminded to check it's intended (there's no deterministic way to validate it).

That said, your comment prompted a bigger rethink (see my reply to your other comment about promoting the coarser case into the check message itself). That removes the need for the warning to carry the full explanation, so the wording became much shorter as a natural consequence.

R/check_tbl_spl_compound_taskid_set.R
annakrystalli: Great suggestion, you pushed me to go further with this. I've promoted the coarser case into the check message itself rather than keeping it hidden behind `show_warnings = TRUE`. Specifically:

- **Success message** now distinguishes match vs coarser: *"…matches the configured `compound_taskid_set`"* when all modelling tasks match, *"…matches or is coarser than the configured `compound_taskid_set`"* when any are coarser.
- **Per-mt details** are appended to the message in the coarser case, e.g. *"mt 2: detected (...) is coarser than configured (...)"*. These surface by default, no flag needed.
- **Warning is retained but simplified** to a brief pointer (*"Coarser-than-configured `compound_taskid_set` detected for modeling task 2. See check message and compound_taskid_set fields for specifics"*). The structured warning remains useful because #338 plans to aggregate check-level warnings into an end-of-run summary across files, and this warning is the hook.

annakrystalli:
Thanks again for the review @LucieContamin! Your comments prompted a rethink that ended up materially improving the PR:

- Coarser-than-configured `compound_taskid_set` is now surfaced in the **check message itself** (with per-mt detected vs configured details) rather than hidden behind `show_warnings = TRUE`.
- The `validation_warning` is retained as a brief pointer (also the hook for #338's end-of-run warning summary).
- Added documentation on `capture_validation_warning()` clarifying that validation warnings never cause PR failure.

See individual replies for details. Ready for another look when you have a moment!

R/check_tbl_spl_compound_taskid_set.R
```diff
@@ -62,36 +62,44 @@ check_tbl_spl_compound_taskid_set <- function(
...
     )
-    if (length(coarser_warnings) == 0L) {
-      coarser_warnings <- NULL
-    }
+  }
+  any_coarser <- length(coarser_info) > 0L
+
+  msg_attribute <- if (!check || any_coarser) {
+    "to single, unique compound task ID set that matches or is
+    coarser than the configured {.var compound_taskid_set}."
```
LucieContamin, line 81: I might be a me problem, but I found the final sentence kind of difficult to  understand, especially the `that matches` part: 

```
All samples in a model task do not conform to single, unique compound task ID set that matches or is coarser than the configured `compound_taskid_set`.
```
I think the "that matches" might be redundant here, no?

```
All samples in a model task do not conform to single, unique compound task ID set or are coarser than the configured `compound_taskid_set`.
```

LucieContamin (approved):
It all looks good, thank you for the update! I like the new output, it's clearer. I just had a small remarks on the phrasing of the error message, but feel free to ignore if you disagree

R/check_tbl_spl_compound_taskid_set.R
annakrystalli: "That matches" is actually doing real work in the sentence, because it's asserting two separate properties:

1. samples conform to a single, unique compound task ID set (internal consistency across sample IDs), and
2. that set matches (or is coarser than) the configured `compound_taskid_set` (compliance with the config).

A table can satisfy (1) while failing (2). For example, samples can be internally consistent but imply a *finer* compound structure than the config allows. The top-level message states the properties checked, and the per-modeling-task `details` appended to it spell out the exact cause in each failure mode (non-unique sets vs finer-than-allowed). Dropping "that matches" would collapse the two properties, so I'd like to keep the current wording.

Issue #357, opened 2026-07-16
Grid-free `check_tbl_values_required` via config-driven coverage validation


annakrystalli:
# Refactor check_tbl_values_required to config-driven, grid-free coverage validation

Parent: #295. Depends on: #355 (routing + distinguishability check).

## Problem

`check_tbl_values_required()` verifies all required combinations were submitted by
building both the required grid (`req`) and the full grid (`full`) per modeling task and
performing repeated joins/anti-joins, including a combinatorial pass over subsets of
optional values. It is the most intricate check in the package and materialises the full
grid, so it carries the same OOM risk and is hard to maintain.

## Behavioural contract to preserve (the hard part)

The refactor MUST reproduce these rules exactly:

- **R1. Required grid fully present.** Per model task, the product of every column's
  *required* values must all appear.
- **R2. Conditional optional completeness.** A task ID may carry both required and
  optional values. Optional values are not obligatory, but if one is submitted it must be
  submitted against the full set of required values in the other columns (submit optional
  `horizon=2` for `location=US` -> now owe `horizon=2` for every required location).
- **R3. Only optional values actually present trigger R2.** Never-submitted optional
  values are never flagged.
- **R4. Optional values interact at every level.** With multiple mixed columns, R2 holds
  for each optional value alone and for every combination of them (the current
  `get_opt_val_combs` / successively-smaller-`m` machinery).
- **R5. Wholly-missing required model task caught (#203).**
- **R6. Applicability / no cross-model-task leakage (#177).** Expected required values are
  restricted to what the model task actually permits.

## Simplification unlocked by later schemas

Later schemas have **no optional `output_type_id`s, only optional output types** (v4
already forces all ids of a submitted output type to be required). Therefore
`output_type_id` is never a mixed column, and we can delete:

- `ignore_optional_output_type()` and `get_required_output_types()`,
- `output_type_id`'s participation in the optional-combination explosion.

After this, the only mixed columns are **task IDs**; `output_type` is all-or-nothing.

## Key insight / proposed design

Everything above can be driven from the config's per-column required/optional sets plus
the submission grouped by optional signature — neither needs the full grid. Per modeling
task (inside an output-type slice, rows assigned via Issue 1):

1. From config, get each column's required set `Rc` and optional set `Oc`.
2. Group the assigned rows by their optional signature (mask required-valued cells, as
   `get_group_rows` already does), for rows assigned to the model task via #355. Distinct
   signatures are bounded by submitted rows.
[...]

Issue #356, opened 2026-07-16
Grid-free `check_tbl_values` via set-membership validation


annakrystalli:
# Refactor check_tbl_values to grid-free set-membership validation

Parent: #295. Depends on: #355 (routing helper + config distinguishability check).

## Problem

`check_tbl_values()` validates that each submitted row is a valid task ID / output type /
output type ID combination by, per output type, building the full accepted-values grid
(`expand_model_out_grid(all_character = TRUE, output_types = <ot>)`) and `left_join`-ing
the data to it; rows with no match (`is.na(valid)`) are invalid. The accepted grid is the
Cartesian product across all modeling tasks and is the primary OOM/slow-join source
(≈47.6% of validation wall time is in the join engine per the #295 profile).

## Key insight

A row is a valid combination iff there exists a modeling task whose per-column allowed
value sets all contain the row's values. Membership in a product never requires building
the product. Concretely, per output-type slice (which the check already splits on):

```
valid_count <- integer(nrow(slice))
for (mt in model_tasks_offering_this_output_type) {
  in_mt <- rep(TRUE, nrow(slice))
  for (col in columns(mt)) in_mt <- in_mt & (slice[[col]] %in% mt$allowed[[col]])
  valid_count <- valid_count + in_mt
}
invalid <- valid_count == 0L
```

Cost O(rows x columns x model_tasks) with hashed `%in%`, peak memory a few
data-length logical vectors, independent of the product size. Fold over model tasks so
all-at-once matrices are never held.

Within an output-type slice, `output_type_id` has a fixed enumerated domain and is a full
participating column (it can be the sole discriminator between two model tasks); for
`sample` rows it is NA-masked as today and left to the structural sample checks.

## Behavioural contract to preserve

- **Invalid value vs invalid combination distinction.** Current output separates "column
  X contains invalid value V" (value in no model task's set for that column) from
  "row contains a valid-but-invalid *combination*" (each value valid somewhere, but no
  single model task contains them together). Both fall out of the same vectors: the union
  membership per column gives invalid values; a `valid_count == 0` row whose every value
  is individually valid is an invalid combination.
- **`error_tbl`** contents and column set (`names(tbl) != "value"`) and the row indices
  reported (with respect to original tbl row order) unchanged.
- Message wording / `capture_check_cnd` output unchanged.
- v3 sample `output_type_id` NA handling and derived-task-id NA handling unchanged.
[...]

Issue #295, opened 2025-12-03
Performance optimization: Adopt Arrow + DuckDB + disk cache for memory-efficient validation


annakrystalli:
## Summary

**Bottlenecks being addressed:**
1. **Memory exhaustion** — Creating expanded grids of valid task ID/output type combinations can produce millions of rows (e.g., 5 task IDs × 50 values = 312.5M combinations), exceeding available RAM
2. **Slow join operations** — Comparing submitted data against these large valid combination grids is computationally expensive
3. **Repeated computation** — The same valid combination grids are rebuilt from scratch for every file in a PR, wasting time in CI
4. **Sequential file processing** — Files in a PR are validated one at a time, not taking advantage of available cores
5. **Duplicate in-memory data** — Validation may require both original and character-coerced versions of submitted data, doubling memory for large files

**Proposed Changes:**

- **Arrow tables** — Replace R data frames with Arrow's columnar format in targeted places (e.g., expanded grids). Columnar storage is more memory-efficient (contiguous columns enable compression and avoid R's per-object overhead). Arrow also provides a multi-threaded Parquet reader for faster file I/O.

- **DuckDB** — Use as the query engine for join operations. DuckDB automatically spills to disk when joins exceed available RAM, preventing memory exhaustion. It integrates seamlessly via dbplyr, so existing dplyr code requires minimal changes.

- **DuckDB views for submitted data** — Register the submitted `tbl` as a DuckDB table once per file validation, with a character-cast view created on-the-fly. Avoids holding two full copies (original + character) in R memory.

- **pins package** — Cache the expanded valid combination grids as Parquet files. When validating multiple files in a PR, the grid only needs to be computed once and can be reused for subsequent files. Particularly valuable in CI where the same PR may trigger multiple validation runs.

- **Parallel processing** — Validate multiple files concurrently across available cores. Combined with cached grids, this dramatically reduces total validation time for PRs with many files.

**Why now:** 
- Arrow 17.0.0 (Aug 2024) added NA handling support in dplyr joins, removing a previous blocker
- Hub integrity checks (#83) require validating many existing files when configs change—impractical without these performance improvements

**Size Estimate:** ~5 weeks across 5 phases. Arrow and DuckDB are already in the dependency tree; only pins is new. Touches grid creation, join operations, and CI/CD workflows.

**Opportunities:**
- 50-70% memory reduction and up to 90% speedup for large PRs
- Parallel file validation and multi-threaded reads in GitHub Actions

[...]

PR #360, opened 2026-07-28
Add validation benchmark with baseline measurements

### LINKED ISSUE #323: Phase 0: Benchmarking — Establish performance baselines
## Goal

Establish robust performance baselines **before any code changes** so we can make informed decisions about which optimizations are worth the added complexity, and track progress through subsequent phases.

As noted in the [review discussion](https://github.com/hubverse-org/hubValidations/issues/295#issuecomment-3785488724), for most existing hubs validation performance isn't currently a major pain point. Starting with benchmarks ensures we're weighing benefits against complexity with concrete data rather than estimates.

## Core Metrics to Track

**Memory:**
- Peak memory usage
- Memory allocated during operations
- Object sizes (grids, submitted data)

**Performance:**
- Execution time (grid creation, joins, full validation pipeline)
- Maximum grid size before OOM

**Tools:** `bench::mark()` with `memory = TRUE`

## Test Cases

**Tiered synthetic configs:**
- **Small** (60 rows) — regression check
- **Medium** (750 rows) — typical hub
- **Large** (3,200 rows) — stress test
- **Extra Large** (10,000 rows) — where improvements matter most

**Real hub configs:**
- FluSight-forecast-hub (or equivalent)
- covid19-forecast-hub (or equivalent)
- SMH (volunteered by @LucieContamin — has submissions with ~25M rows)

## Essential Benchmark Functions

```r
# 1. Grid creation
benchmark_grid_creation <- function() {
  bench::mark(
    tibble = expand_model_out_grid(config, as_arrow_table = FALSE),
    iterations = 5,
    memory = TRUE
  )
}

# 2. Validation pipeline
benchmark_validation <- function() {
  bench::mark(
    current = validate_pr(hub_path),
    iterations = 3,
    memory = TRUE
  )
}
```

## CI Benchmarking

Benchmarks should run both locally and on GitHub Actions to capture performance under real CI conditions (standard runner: 4 vCPUs, 16 GB RAM).

**Triggering strategy:**
- **`workflow_dispatch`** — manual trigger from the Actions tab for ad-hoc runs (baseline establishment, one-off comparisons). Can include input parameters (e.g. which config sizes to test).
- **`/benchmark` PR comment** — trigger benchmarks against a PR branch by posting a `/benchmark` comment. Results posted back to the PR as a comment. This requires handling the `issue_comment` event trigger (explicit PR branch checkout), but is manageable.

**Workflow outputs:**
- Benchmark results saved as artifacts
- Summary posted as a PR comment (when triggered from a PR)
- Baseline CSVs committed to the repo for comparison

## Deliverables

- [ ] Benchmark functions for grid creation, joins, and full validation pipeline
- [ ] GitHub Actions workflow with `workflow_dispatch` and `/benchmark` comment triggers
- [ ] Baseline benchmark results saved as CSV (local + CI)
- [ ] Identify where real bottlenecks are (grid creation vs joins vs other)
- [ ] Characterise memory profile of `expand.grid()` for different config sizes
- [ ] Document which hub configs are most affected

## Success Criteria

- Clear understanding of current performance characteristics
- Saved baseline data that subsequent phases can compare against
- Informed view on which optimizations will have the most impact
- CI benchmarking workflow operational

## Estimated effort
2–3 days

annakrystalli:
Adds `_benchmark/`, which measures how long hub validation takes and how much
memory it needs, so the grid-free rewrite in #355, #356 and #357 can be judged on
numbers rather than estimates. No package code changes; `_benchmark/` is
`.Rbuildignore`d.

Resolves #323.

## What it does

Test hubs are generated from the config ruarai shared on #295 because validating
their submissions was slow. Two settings vary independently:

- **size** — how much data is submitted, and how many rows the config's valid value
  combinations grid has. S/M/L grow both together, so L is their hub exactly.
  G1–G3 hold the submitted data at ~65k rows and grow only the config, so any
  change in memory is caused by the config rather than by the data.
- **shape** — how many model tasks the config is split into (`acefa` = 1, then
  `mt3`, `mt7`, …). The split shares one set of values out between them, so the grid
  stays the same size and the cost of having more model tasks is separable from the
  cost of a bigger config.

Two modes: `submission` validates a whole file and profiles it; `peak` runs each
check in its own process and reads peak memory from the OS, which is the only figure
that also counts memory Arrow holds outside R.

## Baseline

Full numbers in `_benchmark/prof-summary.md`:

| | |
|---|---|
| one file at the reported size (13M rows) | **~29 min**, ~11 GiB, 0 errors |
| of which `check_tbl_values_required` | **91%** |
| of which matching rows against the grid | **89%** of self time |
| 65k rows against a 39M-row grid | **~10 GB** |
| grid ×3.3 with submitted data held still → memory | **×1.5–3.2** |
| the two checks that don't assign rows to model tasks | **×1.0** |

Two things worth knowing before starting the rewrite:

1. **`check_tbl_values_required` (#357) carries the cost, not `check_tbl_values`
   (#356)** — 91% against under 1% at size L. Its expense is working through each
   model task's optional-value combinations, so it gets *faster* when the config is
   split across more model tasks (15.9 s → 6.2 s from 1 to 7 at size M). This
   config's single all-optional model task is close to the worst case, so the size L
   figures are an upper bound rather than typical.
2. **~20% of a mid-size run is in the sample checks**, which #355 defers. They decide
   which model task a row belongs to the same way, so the same helper addresses them.
   That share falls to ~6% at size L.

## Reviewing

This isn't package code, so a line-by-line review isn't needed and there's no need to
run anything. The useful things to cast an eye over:

- **`_benchmark/README.md`** — does the framework measure the right things, and is it
  clear enough that someone else could pick it up? Reference material first, design
  detail in the appendix.
- **`_benchmark/prof-summary.md`** and the two CSVs — do the baseline numbers and the
  conclusions drawn from them look sound?

Treat time differences under ~5% and memory differences under ~20% as noise; peak
memory in particular varies that much between runs.

Optionally, to see it work:

```sh
# ~30 s
HUBVALIDATIONS_BENCHMARK_SIZES=S,M HUBVALIDATIONS_BENCHMARK_MODE=submission \
  Rscript _benchmark/run-benchmark.R

# ~3 min, per-check time and memory
HUBVALIDATIONS_BENCHMARK_SIZES=M HUBVALIDATIONS_BENCHMARK_MODE=peak \
  Rscript _benchmark/run-benchmark.R
```

Test hubs are generated on first use, so nothing large is checked in. Size L takes
~30 minutes and G2 ~45; both are in the committed baselines already.


micokoch (approved):
Overall this is clear and I was able to follow as well as I can understand what is going on. I asked Claude to review the CSV files and the PR itself and it had some helpful comments which I include below:

Summary: Framework looks sound — good design (data-vs-config isolation, OS-level RSS for Arrow, flat control checks), and the table numbers reconcile with the CSVs. A few small fixes, only #1 matters before merge.

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,115 @@
...
+
+What validation costs today: the numbers #355–#357 have to beat. `results.csv` and
+`peak-results.csv` hold every run; this file is the short version. Only add to it
+when a run moves a cost, removes one, or turns up something new.
+
+Machine: Darwin arm64, R 4.5.2.
+
+## Headlines
+
+1. **One file at the size that was reported takes ~29 minutes and ~11 GiB** (size L:
```
micokoch: 1. Size-L memory headline is inconsistent (`prof-summary.md`). Headline says "~11 GiB", but the table says 17.3 GiB (R held) and the prose says peak 15.4 GiB (range 10.7–15.4). Pick the figure people should budget for (≈15 GiB) and note the two memory numbers are different metrics (R gc vs OS RSS) — otherwise 17.3 > 15.4 reads as a bug.

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,115 @@
...
+   13M submitted rows, 3.9M valid value combinations grid rows, one model task).
+2. **`check_tbl_values_required` accounts for 91% of that time**, and matching rows
+   against the valid value grid accounts for 89% of it. Everything else put
+   together is under 10%.
+3. **Memory grows with the size of the valid value grid, not with how much data was
+   submitted.** Hold the submission at 65k rows and triple the grid, and memory
+   roughly triples with it: validating 65k rows costs about 10 GB when the grid has
+   39M rows.
```
micokoch: 2. "memory roughly triples" overstates (headline 3). Body says 1.5–3.2x and whole-set peak only grows ~1.5x. Conclusion (memory tracks config) is fine; suggest "grows 1.5–3.2x".

_benchmark/README.md
```diff
@@ -0,0 +1,265 @@
...
+grid has. Called the **valid value grid** from here on.
+
+| size | submitted rows | valid value grid rows |
+|---|---:|---:|
+| S | 18,000 | 25,200 |
+| M | 460,000 | 237,510 |
+| L | 13,000,000 | 3,887,250 |
+| G1 | 65,000 | 11,661,750 |
+| G2 | 65,000 | 38,872,500 |
+| G3 | 65,000 | 77,745,000 |
```
micokoch, line 69: 3. G3 defined but never measured. README describes G3 (77.7M grid) and tells you to run "G1, G2 and G3", but there's no G3 in the CSV or summary. Add a line on why (OOM? skipped?).

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,115 @@
...
+4. **Working out which model task each row belongs to already costs more when there
+   are more of them**: ~3.3x slower going from 1 to 7. So the new approach checking
+   rows against each model task in turn is not automatically worse than what we have.
+   Keeping a separate marker per row for every model task would be.
+5. **`check_tbl_values_required` gets ~2.7x *faster* with 7 model tasks** than with
```
micokoch: 4. (minor) Headline 5 "~2.7x" is really 2.55x → ~2.6x. Noise-level.

micokoch (approved):

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
+# The list of checks the benchmark measures, how validate_model_data() calls each
+# one, and what each needs to be given. Used by run-benchmark.R and by
+# run-one-check.R, which runs a single check on its own so its memory use can be
+# measured cleanly.
+
+# `scope` says whether the grid-free work is expected to change a check, and it is
+# the main reason this list exists. These test hubs submit samples, so a fair chunk
+# of every run is spent in the sample checks. Without the label, that time would
+# make a successful change look disappointing, because it was never going to move.
```
LucieContamin: I don't understand that paragraph, sorry. Would it be possible to rephrase, please?

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
...
+#
+# - `in`        — rewritten by #355-#357.
+# - `follow-up` — the sample checks. #355 leaves them out, but they work out which
+#   model task a row belongs to in exactly the same way it is replacing: build the
+#   valid value grid for each model task, then match the data against it.
+#   check_tbl_spl_mt_unique does that directly, and spl_hash_tbl()
+#   (utils-samples.R) and get_tbl_compound_taskid_set() (compound_taskid-utils.R)
+#   do it too. So their cost can be dealt with by reusing #355's new helper rather
+#   than by anything new, which is why they are tracked separately instead of
+#   written off.
```
LucieContamin: Not sure I understand the last sentence, sorry

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
...
+# - `other`     — neither, e.g. check_tbl_rows_unique.
+#
+# The labels describe the plan as it stands for #355-#357. Once that work lands they
+# describe what it did touch, so they should not be edited in place: rows already
+# saved in the CSVs would stop meaning the same thing as new ones with the same
+# label.
```
LucieContamin: I am not clear on that part neither, sorry

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
...
+# process measuring one check reads only what that check actually uses. Reading the
+# other version as well would add all of its memory to the figure being blamed on
+# the check.
+CHECKS <- list(
+  # Not a check itself. It is the shared code underneath check_tbl_value_col and
+  # check_tbl_value_col_ascending that works out which model task each row belongs
+  # to, measured on its own because #355 replaces it. Its two real callers each
+  # name one output type; leaving that unset here covers all of them at once. On
+  # these test hubs that is the same thing, since they only use samples, and on a
+  # hub with more it is deliberately the worst case.
```
LucieContamin: Same, sorry I don't know if it's me if I am missing something but I am having trouble understanding the documentation here

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
...
+  # change them, and they are too quick to register at any size.
+  "check_valid_round_id_col",
+  "check_tbl_unique_round_id",
+  "check_tbl_match_round_id",
+  "check_tbl_colnames",
+  "check_tbl_col_types",
+  # These do nothing on these test hubs, which have no derived task IDs and no
+  # probability-mass output type.
+  "check_tbl_derived_task_id_vals",
+  "check_tbl_value_col_sum1"
```
LucieContamin: Same, I think I got the idea, but the documentation is not clear to me

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
...
+  # number of columns rather than the number of rows. Grid-free validation cannot
+  # change them, and they are too quick to register at any size.
+  "check_valid_round_id_col",
+  "check_tbl_unique_round_id",
+  "check_tbl_match_round_id",
+  "check_tbl_colnames",
+  "check_tbl_col_types",
+  # These do nothing on these test hubs, which have no derived task IDs and no
+  # probability-mass output type.
+  "check_tbl_derived_task_id_vals",
```
LucieContamin: Should we have a test hubs with derived task id to have a good idea of the impact of that part, too?

_benchmark/check-calls.R
```diff
@@ -0,0 +1,331 @@
...
+    stop("No call defined for check '", name, "'", call. = FALSE)
+  )
+}
+
+# validate_model_data() reads the submission twice: once with everything as text,
+# and once with proper column types. For that second read it uses the types from
+# the hub config for csv files, and the types stored in the file itself otherwise.
+# Copying that split matters, because reading a parquet test hub with the config's
+# types would measure the value-column check against types it never actually
+# receives in practice.
```
LucieContamin: Does that mean the value-column check is never happening?

_benchmark/make-hub.R
```diff
@@ -0,0 +1,772 @@
...
+# which means a test hub that lost the gap would make the new code look better
+# than it is. The size list below says which sizes keep it and which change one
+# side on purpose.
+#
+# Sourced by run-benchmark.R; also usable on its own:
+#   Rscript -e 'source("_benchmark/make-hub.R"); make_benchmark_hub("S")'
+
+# Size definitions ------------------------------------------------------------
+
+# Each size is a size. The `cfg_*` entries set how many values the config allows,
```
LucieContamin: Not sure what the first sentence means.

_benchmark/make-hub.R
```diff
@@ -0,0 +1,772 @@
...
+# which decides how big the valid value grid becomes. The rest set how much
+# data is actually submitted.
+#
+# The L size is ruarai's config exactly, which assert_faithful_to_acefa() checks.
+# Every other size is the same shape, scaled up or down.
+#
+# Submitted rows = units * samples * (2 * horizons used + 2), from
```
LucieContamin: Why "+2"?

_benchmark/make-hub.R
```diff
@@ -0,0 +1,772 @@
...
+#
+# - S, M and L grow the config and the submission together, like a real hub.
+# - G1 to G3 grow only the config, so any extra memory has to be the valid value
+#   grid. This is where today's cost sits: memory grows in step with that grid.
+# - the `mt<N>` variants add model tasks while the config and the submission stay
+#   the same size. This is where the replacement could get slow instead. The plan
+#   is to test rows against one model task at a time; if it kept a separate
+#   true/false marker per row for every model task, memory would grow with rows
+#   times model tasks. #356 says not to do that, and this is how we would notice.
+#   It needs a size with a lot of data to show up.
```
LucieContamin: Does that mean we are not going to test a round with multiple model tasks included in the same model output file?

_benchmark/make-hub.R
```diff
@@ -0,0 +1,772 @@
...
+      stringsAsFactors = FALSE
+    )
+  )
+}
+
+# The date, location and disease combinations a submission covers. Their file has
+# 25 of them, spread over 9 locations and 6 dates, and spread unevenly: each
+# location has 1 or 2 dates, and each of those has 1 to 3 diseases. The dates
+# differ because every location and disease has its own last day of reliable data.
+#
```
LucieContamin, line 377: All the submission information is not obvious to me from the config files. I am not sure I understand what is expected from the config task.json to be honest. For example, are all origin date expected for all rounds? I think it will be good to clarify the task.json if we want to use it as testing, to avoid confusion

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,124 @@
+# Baseline
+
+What validation costs today: the numbers #355–#357 have to beat. `results.csv` and
```
LucieContamin: Do you mean have to be beaten?

LucieContamin (changes requested):
Thank you for the update. I don't know if it's only me but I really struggle with the documentation. A lot of phrase were not clear to me, so I am not sure I understand the process properly here. 
Also, I wonder why are we using a specific hub for the benchmark and not different hubs that use the tasks.json in very different way to cover more possibilities?

_benchmark/check-calls.R
annakrystalli: Sorry about this one — the `scope` label was Claude's invention, and on closer inspection it wasn't doing anything useful, so I've removed it along with this paragraph (98dfa35).

_benchmark/check-calls.R
annakrystalli: Gone with the rest of the `scope` labels in 98dfa35.

_benchmark/check-calls.R
annakrystalli: Also gone in 98dfa35. This was a warning never to correct the labels later, which was really a sign the field shouldn't have been there at all.

_benchmark/check-calls.R
annakrystalli: Not you at all. Rewritten in 98dfa35: `tbl` and `compound_taskid_set` are now documented as a list in the header rather than a paragraph, and this entry is shorter. The part that matters is that it isn't a check — it's the shared code that works out which model task each row belongs to, measured on its own because #355 replaces it.

_benchmark/check-calls.R
annakrystalli: Rewritten in 98dfa35. It now says up front that `validate_model_data()` runs more checks than the benchmark times, and these are the ones deliberately left out, with a reason against each group rather than one sentence covering several.

_benchmark/check-calls.R
annakrystalli: I don't think it would tell us anything. Derived task IDs are set to NA before the grid is expanded (`expand_model_out_grid.R:241`), so they never multiply into it — adding one could only make the grid *smaller*, and none of the three things the benchmark varies would move. `config_grid_rows()` skips them for the same reason.

`check_tbl_derived_task_id_vals` is the only check specific to them, and it returns straight away when a config has none, which is why it's on the deliberately-not-measured list rather than being measured at zero.

_benchmark/check-calls.R
annakrystalli: No, it runs normally — sorry, that comment read much worse than I intended. It was only about *which* of the two versions of the submission it gets handed. `validate_model_data()` reads the file twice, once all character and once with real column types, and gives `check_tbl_value_col` the typed one. The benchmark copies how those types are chosen so the check sees the same types here as it would in a real run. Rewritten in 98dfa35.

_benchmark/make-hub.R
annakrystalli: Because it didn't mean anything — it read "Each tier is a size" until Claude renamed tiers to sizes and turned it into nonsense. Rewritten in 98dfa35: it now says what a size actually is, names `units` and `samples`, and explains the row formula.

_benchmark/make-hub.R
annakrystalli: The two peak targets. `case_incidence` and `case_incidence_smoothed` are submitted at every horizon, which is the `2 * horizons used`, while `peak_case_incidence` and `peak_day_of_year` appear once each — hence the `+ 2`. Now explained where the formula is (98dfa35).

See also https://github.com/hubverse-org/hubValidations/issues/295#issuecomment-5105572663, where we're asking the hub's authors about their horizon encoding for the peak targets.

_benchmark/make-hub.R
annakrystalli: We do — that's exactly what the `mt<N>` shapes are. `mt3` splits the config into 3 model tasks in the same round, with one submission file covering all of them.

And it matters for more than I'd written. The grid is built as one sub-grid per model task and validation maps over them, so more model tasks means more passes through the same build-and-join work — a cost in the code as it stands, not just a risk in the replacement. You can see it in `prof-summary.md`: the sample checks take 3.3x longer at 7 model tasks than at 1, on identical data. Validation maps over output type too, which is why these hubs submitting a single output type is useful rather than a shortcoming: every pass is the same shape, so the model task count varies on its own.

I've rewritten that comment and the README's Shape section to say all of this (18eb721).

_benchmark/make-hub.R
annakrystalli: Nothing is expected — every task ID value in this config is optional. So all 365 origin dates are permitted in every round and none are required, which is exactly why the grid is so large. The submission then uses 6 of them.

On clarifying the config: it is confusing, and not only to you. We've asked the hub's authors what they intended, and cc'd you: https://github.com/hubverse-org/hubValidations/issues/295#issuecomment-5105572663. But whatever they come back with won't change the size of the submission, the size of the valid value grid, or the number of model tasks — so it doesn't change what the benchmark measures or how it's set up. The config is just the raw material for producing a submission and a grid at a chosen size.

_benchmark/prof-summary.md
annakrystalli: Yes. Reworded to "These are the numbers #355–#357 need to improve on" (98dfa35).

_benchmark/check-calls.R
```diff
@@ -0,0 +1,315 @@
...
+  check_table <- utils::read.csv(table_path)
+  expected <- check_table$check.fun[
+    check_table$parent.fun == "validate_model_data" & !check_table$optional
+  ]
+  measured <- vapply(CHECKS, \(x) x$name, character(1))
+  missing <- setdiff(expected, c(measured, DELIBERATELY_UNMEASURED))
+  if (length(missing) > 0L) {
+    cat(
+      "note: validate_model_data() runs checks the benchmark does not measure:\n  ",
+      paste(missing, collapse = ", "),
```
LucieContamin, line 101: Minor issue: Is it me or the phrase is missing something? Should we rephrase it, maybe:

```suggestion
      "note: validate_model_data() runs checks that the benchmark does not measure:\n  ",
      paste(missing, collapse = ", "),
```

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,134 @@
...
+`validate_submission()` on one submission file, from `results.csv`. "Valid value grid
+rows" is how many rows that grid has for the round.
+
+| size | submitted rows | valid value grid rows | time | R heap peak | errors |
+|---|---:|---:|---:|---:|---:|
+| S | 18,000 | 25,200 | 2.5 s | 188 MB | 0 |
+| M | 460,000 | 237,510 | 25.6 s | 735 MB | 0 |
+| L | 13,000,000 | 3,887,250 | 1,769 s | 17,751 MB | 0 |
+
+**Two different memory numbers, don't compare them.** The table's "R heap peak" is R's
```
LucieContamin: Which table is "R heap peak"?

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,134 @@
...
+own high-water counter (`gc()` max-used), which read 17.3 GiB at size L. The number that
+decides whether a run fits on a machine is the whole process's resident set, from
+`/usr/bin/time -l`, which read 10.5 GiB for this run. R's counter can exceed the
+resident set because it tracks everything R ever had allocated at once, while pages get
+freed and reused, so 17.3 > 10.5 is not a contradiction. Budget on the process figure.
```
LucieContamin: `Budget on the process figure.` I am not sure I understand that phrase

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,134 @@
...
+
+**Expect noise.** Across four runs, size L took 1,694-1,813 s, so treat time differences
+under about 5% as nothing. The process peak varied more, 10.5 to 15.4 GiB, since it
```
LucieContamin, line 60:
```suggestion
**Expect noise.** Across four runs, size L took 1,694-1,813 s, so we treat time differences
under about 5% as nothing. The process peak varied more, 10.5 to 15.4 GiB, since it
```

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,134 @@
...
+| `check_tbl_rows_unique` | 2.8 |
+
+Matching rows against the valid value grid accounts for 88.9% of the time on
+its own. At size M that same check is 64% rather than 91%, and the sample checks are
+~22% rather than ~6%. So the bigger the hub, the stronger the case for #357.
+
+## Cost as the config grows
+
+The submission is held at 65,000 rows and only the config grows, so anything that moves
+here is down to the valid value grid. Peak memory and elapsed time per check.
```
LucieContamin: The last phrase is missing some verb no? or is it a description of the table below?

_benchmark/prof-summary.md
```diff
@@ -0,0 +1,134 @@
...
+and 14 GB at G3 — they are converging on what this machine will hand out rather than on
+what the work needs, so read the memory column as a floor on the real appetite, not a
+measurement of it.
+
+The last two rows are the control. Neither check has to work out which model task a row
+belongs to, so neither is affected by the size of the grid: both sit at ~200-290 MB and
+under a second across a grid that grew nearly sevenfold.
+
+What to expect of the other rows once #355-#357 land: they should **drop**, and the gaps
+between the G columns should **narrow**. Not to the controls' floor, though. A config
```
LucieContamin, line 108:
```suggestion
What to expect of the other rows once #355-#357 land: they should **drop**, and the gaps
between the G columns should **narrow**, not to the controls' floor, though. A config
```

_benchmark/README.md
```diff
@@ -0,0 +1,287 @@
...
+  G3 should still cost more than G1 — just far less than it does today.
+
+For a check that never touches the grid, `check_tbl_rows_unique` shows what the floor
+looks like: ~290 MB and under a second at all three sizes.
+
+Read `elapsed_s` alongside `peak_rss_mb`, and prefer it. Across those three the grid
+grows 6.7x and the times grow with it, but the memory figures flatten as they approach
+what the machine will give. Everything affected lands between 11 and 14 GB at G3, so
+memory understates the effect at the top end. Peak memory also moves by around 20%
+between runs, so only believe a memory change larger than that.
```
LucieContamin: Not sure I understand that phrase, sorry

_benchmark/run-benchmark.R
```diff
@@ -0,0 +1,532 @@
+# Times and measures the memory of validation against the test hubs, so the
+# grid-free work in #355 / #356 / #357 can be judged on numbers rather than hope.
```
LucieContamin, line 2: I am sure I understand the "hope" part

_benchmark/run-benchmark.R
```diff
@@ -0,0 +1,532 @@
...
+  strsplit(env_default("HUBVALIDATIONS_BENCHMARK_VARIANTS", "acefa"), ",")[[1]]
+)
+hubs_dir <- env_default(
+  "HUBVALIDATIONS_BENCHMARK_HUBS",
+  file.path(benchmark_dir, "hubs")
+)
+# A best effort rather than a guarantee. R only notices the limit between steps of
+# its own work, so a check stuck inside one long operation in compiled code (a big
+# join, which is what the large sizes risk) will run past it. Wrap the whole thing
+# in `timeout` if you need a hard stop.
```
LucieContamin:
```suggestion
# R only notices the limit between steps of
# its own work, so a check stuck inside one long operation in compiled code (a big
# join, which is what the large sizes risk) will run past it. Wrap the whole thing
# in `timeout` if you need a hard stop.
```

_benchmark/run-benchmark.R
```diff
@@ -0,0 +1,532 @@
...
+  result_line <- grep("^BENCHMARK_RESULT", output, value = TRUE)
+  parse_field <- function(field, default) {
+    if (length(result_line) != 1L) {
+      return(default)
+    }
+    sub(paste0(".*", field, "=([^ ]+).*"), "\\1", result_line)
+  }
+  elapsed_s <- as.numeric(parse_field("elapsed_s", NA_character_))
+  # A check that fails is still a successful run of the script, so the exit code
+  # alone cannot tell us. Recording what the check returned means a check that
```
LucieContamin: What do you mean by the exit code alone cannot tell us? If the check fails or not?

LucieContamin (commented):
Thank you for the update and the explanation, it makes more sense now!  Thank you for tagging me on the ongoing issue about the config files, I am curious about it! 

I think there is still some "Claude language" part that are difficult to read, at least for me in the documents, but globally it looks all ok and makes sense to me now!
I made some minors suggestions or remarks, feel free to ignore.

LucieContamin (approved):

_benchmark/prof-summary.md
annakrystalli: Ambiguous with two tables nearby — fixed. Now names the column and points at the table above: "The `R heap peak` column in the table above is R's own high-water counter".

_benchmark/prof-summary.md
annakrystalli: That was jargon I'd invented and never defined. Replaced with what it actually means: "When working out how much memory a machine needs, use the resident set figure, not this column."

_benchmark/prof-summary.md
annakrystalli: Both — it was a fragment describing the table. Now a sentence: "Each cell gives the peak memory and the elapsed time for one check at one size."

_benchmark/README.md
annakrystalli: Two problems in that paragraph: "prefer it" did not say prefer which, and "memory understates the effect" asserted something without explaining why. Rewritten to say the reason — the memory figures level off because they hit what the machine will hand out, not because the work stops growing, so at G3 they make the effect look smaller than it is.

_benchmark/run-benchmark.R
annakrystalli: Exactly that, and the sentence never finished the thought. Now: "a zero exit code only tells us the script ran, not that the check passed."

PR #362, opened 2026-07-29
Handle a compound_taskid_set of []

annakrystalli:
Resolves #361.

A `compound_taskid_set` of `[]` declares that every task ID is sampled jointly. It
passes config validation, but `read_config()` returns it as an empty list rather than
an empty character vector, and an empty list cannot index a column. So
`check_tbl_spl_compound_taskid_set()` stopped with an exec error, and
`submission_tmpl()` failed with `invalid subscript type 'list'`.

## The fix

The issue suggested guarding the assignment that breaks first. The fix in this PR
instead normalises the value where the config is read, so every consumer gets a
character vector. That matters because the empty list reaches more than one place, as
the rest of this description shows.

Three further problems only became visible once an empty set was a value that reached
those consumers:

- `get_tbl_compound_taskid_set()` dropped it when compacting, because
  `purrr::compact()` removes anything of length zero. A modeling task whose samples
  are correctly jointly sampled was indistinguishable in the output from one with no
  samples at all, which contradicted the function's own documentation. It now drops
  only `NULL`.
- Building sample indices joined on no columns at all, which dplyr deprecated in
  favour of `cross_join()`. This is why `submission_tmpl()` warned once the first fix
  was in.
- The "finer than the hub accepts" message ended in "should be one of ." when there
  was nothing to list.

## Message wording

That last message now describes what the hub accepts rather than what its config
allows, since the config is a statement of what the hub wants in submissions rather
than a list of permissions. **This applies to every hub, not only those declaring
`[]`**, and two existing snapshots change accordingly:

> Detected a finer `compound_taskid_set` than the hub accepts. "horizon" identified as
> compound task ID in file but not accepted by the hub. Compound task IDs should be one
> of "reference_date" and "location".

For an empty set the last sentence instead reads "The hub expects no compound task IDs
and response dependence across all task IDs", using the term
[the sample docs use](https://docs.hubverse.io/en/latest/user-guide/sample-output-type.html).

## A docs gap this turns up

The sample docs state the general rule — task IDs not in the `compound_taskid_set` are
sampled jointly — from which `[]` follows unambiguously: none are in the set, so every
task ID is sampled jointly. The empty case is never named or exemplified there, and an
absent `compound_taskid_set` means the opposite, so the two are easy to conflate. Raised
as hubverse-org/hubDocs#507.

## Testing

- A jointly sampled submission passes against a `[]` config and is reported as
  *matching* the configured set; the file as shipped, which keeps each sample to one
  location, fails as finer.
- `validate_model_data()` end to end on a `[]` hub, asserting no errors. The empty set
  has to reach `spl_compound_tid`, `spl_non_compound_tid` and `spl_n` unchanged for
  them to read it as "every task ID is sampled jointly", and nothing else pinned that.
- `submission_tmpl()` returns one sample covering every task ID value combination, with
  no warning.

Adds config fixture `testdata/configs/tasks-samples-empty-cts.json`.

### DIFF
--- a/tests/testthat/test-check_tbl_spl_compound_taskid_set.R
+++ b/tests/testthat/test-check_tbl_spl_compound_taskid_set.R
@@ -299,3 +299,63 @@ test_that("Ignoring derived_task_ids in check_tbl_spl_compound_taskid_set works"
     )
   )
 })
+
+test_that("check_tbl_spl_compound_taskid_set works with an empty compound_taskid_set", {
+  hub_path <- empty_cts_hub()
+  file_path <- "flu-base/2022-10-22-flu-base.csv"
+  round_id <- "2022-10-22"
+  tbl <- read_model_out_file(file_path, hub_path, coerce_types = "chr")
+  tbl <- tbl[tbl$output_type == "sample", ]
+  task_ids <- setdiff(names(tbl), c(hubUtils::std_colnames, "value"))
+
+  # Give every sample one row per task ID value combination, so no task ID holds a
+  # single value within a sample and the detected set is empty, as configured.
+  joint <- tbl |>
+    dplyr::group_by(dplyr::pick(dplyr::all_of(task_ids))) |>
+    dplyr::mutate(output_type_id = as.character(dplyr::row_number())) |>
+    dplyr::ungroup()
+
+  expect_snapshot(
+    check_tbl_spl_compound_taskid_set(joint, round_id, file_path, hub_path)
+  )
+  config_tasks <- read_config(hub_path, "tasks")
+  expect_equal(
+    get_tbl_compound_taskid_set(joint, config_tasks, round_id, compact = FALSE),
+    list(`1` = NULL, `2` = character(0))
+  )
+  # Compacting drops the modeling task without samples but keeps the empty detected
+  # set, which is an answer rather than an absence.
+  expect_equal(
+    get_tbl_compound_taskid_set(joint, config_tasks, round_id),
+    list(`2` = character(0))
+  )
+
+  # The file as submitted keeps each sample to one location, which is a finer set
+  # than a hub expecting none.
+  expect_snapshot(
+    check_tbl_spl_compound_taskid_set(tbl, round_id, file_path, hub_path)
+  )
+})
+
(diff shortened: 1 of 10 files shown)


tests/testthat/test-check_tbl_spl_compound_taskid_set.R
```diff
@@ -299,3 +299,63 @@ test_that("Ignoring derived_task_ids in check_tbl_spl_compound_taskid_set works"
...
+    dplyr::mutate(output_type_id = as.character(dplyr::row_number())) |>
+    dplyr::ungroup()
+  utils::write.csv(
+    rbind(tbl[tbl$output_type != "sample", ], joint),
+    fs::path(hub_path, "model-output", file_path),
+    row.names = FALSE
+  )
+
+  # The empty set has to reach the other sample checks unchanged for them to read it
+  # as "every task ID is sampled jointly" rather than as no configuration at all.
```
lshandross: Could you clarify what this is saying?

lshandross (approved):
Just one clarifying question from me, and an additional request to remove the artificial line breaks in the PR description so it's easier to read and doesn't waste vertical space. But I'm comfortable giving you the approval as is.

tests/testthat/test-check_tbl_spl_compound_taskid_set.R
annakrystalli: Sorry — two problems with it. It was too compressed, and it was in the wrong place: it explains why the test exists but sat above the last two lines, after all the fixture setup.

What it meant: the other sample checks (`spl_compound_tid`, `spl_non_compound_tid`, `spl_n`) all read the compound task ID set. An empty one is easy to lose on the way to them — dropped when compacting, or arriving as `NULL` — and either way they see nothing configured and read it as an *absent* set, which asks for the opposite sampling structure ([the docs explain the two](https://docs.hubverse.io/en/latest/user-guide/sample-output-type.html)). Silently, since the submission would still validate.

Not hypothetical: `get_tbl_compound_taskid_set()` was dropping it, which is the third bullet in the description. So this test calls `validate_model_data()` rather than the check on its own, to pin that the empty set reaches them intact.

Moved to the top of the test and rewritten to say that.

Issue #368, opened 2026-08-05
Grid-free sample checks: assign sample rows by set membership


annakrystalli:
Three places still assign submitted rows to modeling tasks by expanding a grid and joining the data to it, all on the sample path. `assign_tbl_to_model_task()`, added in #355, does that job by testing each column against the values the modeling task allows, without building the grid at all.

## Direct replacements

`check_tbl_spl_mt_unique()` (`R/check_tbl_spl_mt_unique.R:45`) expands a sample grid per modeling task and `semi_join()`s the submission to each to collect the `output_type_id`s that modeling task claims. That is assignment and nothing else.

`get_tbl_compound_taskid_set()` (`R/compound_taskid-utils.R:58`) expands the same grids and `inner_join()`s to them, dropping `output_type_id` from the grid side first. Also assignment and nothing else.

## Needs more than assignment

`utils-samples.R:44` expands with `include_sample_ids = TRUE` and a `compound_taskid_set`, renames the grid's `output_type_id` to `compound_idx`, and joins so that each row carries that compound index. Here the grid is not only deciding which modeling task a row belongs to, it is generating the grouping of task ID values that the compound task ID set implies. Moving this off the grid means computing that index directly from the compound task ID columns instead of reading it off an expansion, so this is the substantial part of the issue rather than a swap.

## Why it is worth doing

Sample checks are around 20% of a mid-size validation run, and the grid's cost grows with what the config permits rather than with what was submitted, so the checks get more expensive as a hub's config grows even when submissions do not.

## Out of scope

`check_tbl_values()` and `check_tbl_values_required()` have their own issues. `submission_tmpl()` genuinely needs the full grid, since generating every valid combination is the point of it.

PR #369, opened 2026-08-05
Assign submitted rows to modeling tasks without expanding the grid

annakrystalli:
Resolves #355.

## What this changes

`match_tbl_to_model_task()` finds the modeling task that each submitted row belongs to.

Before, it built the grid of every value combination the config allows, then joined the data to that grid. Now it checks each column separately. A row belongs to a modeling task when every one of its values is allowed by that modeling task for that column.

The grid is never built. Memory now depends on the size of the submitted file, not on the size of the config.

## Measurements

G1, G2 and G3 all submit about 65,000 rows. They differ only in what the config allows, which gives grids of 11.7M, 38.9M and 77.7M rows. The table shows peak memory and time, before and after.

| check | size | before | after |
|---|---|---|---|
| `match_tbl_to_model_task` | G1 | 3,568 MB / 8.59s | 221 MB / 0.05s |
| | G2 | 9,783 MB / 27.67s | 210 MB / 0.04s |
| | G3 | 11,452 MB / 60.14s | 211 MB / 0.05s |
| `check_tbl_value_col` | G1 | 3,579 MB / 8.71s | 228 MB / 0.08s |
| | G2 | 9,925 MB / 28.70s | 229 MB / 0.08s |
| | G3 | 13,424 MB / 59.64s | 222 MB / 0.08s |

No other check changed by more than the normal variation between runs. Rounds with more modeling tasks are also faster than before. At size M with 7 modeling tasks, matching goes from 1.07s to 0.71s, and `check_tbl_value_col()` from 1.20s to 0.80s.

## Guide to the files

**`R/assign_tbl_to_model_task.R`** is new and does the work.

- `get_config_mt_value_sets()` reads the config once. For each modeling task, it returns the values allowed in each column.
- `assign_mt_rows()` selects the rows. It first uses `match()` on `output_type_id` to find the position of each row in the config list for its output type. A row with no position is not accepted by this modeling task. It then checks each task ID column with `%in%`.

One point to note while reading this. The submitted data is never converted. `%in%` calls `match()`, and `match()` converts both sides to a common type. That type is character if either side is character. The submitted data is already character, so only the config values are converted. The config holds a few values and the data holds millions of rows, so this is important. This is also why `tbl` must be character.

**`R/match_tbl_to_model_task.R`** has the same arguments as before and calls the new function. `join_tbl_to_model_task()` is unchanged. `check_tbl_values_required()` still uses it, and #357 will replace it in a separate PR.

**`R/expand_model_out_grid.R`**: the part that reads the config moved to `extract_round_property_values()`. Both `expand_model_out_grid()` and `get_config_mt_value_sets()` call it, so they always read the config in the same way. The part that builds the grid is unchanged.

**`R/validate_model_data.R`** now gives `check_tbl_value_col()` the character table `tbl_chr` instead of the typed table.

## Why `check_tbl_value_col()` now uses `tbl_chr`

It was the only data check given the typed table, and it converted every column except `value` to character itself. Each run therefore built the character version twice. The exception for `value` was pointless, because the check converts `value` to the type the config declares anyway.

It now uses the shared character table. This is not a speed change (about 5 ms on 460,000 rows); a run simply holds one character copy instead of two. Output is unchanged for the same input.

## Changes in behaviour

These four are listed in NEWS under Breaking Changes.

**`all_character` is removed.** It controlled whether the grid was built as character, so that the grid and the data had the same types and could be joined. There is no join now, and `tbl` is always character, so the argument has nothing left to control.

**Rows are returned in the order they were submitted.** Before, the join returned them in config order. `check_tbl_value_col_ascending()` needs that order, because it calls `diff(value)` on the rows in the order it receives them. Sorting the `output_type_id` column does not work, because `pmf` categories are not always in alphabetical order (#78). The new argument `order_by_config = TRUE` returns rows in config order. `check_tbl_value_col_ascending()` now asks for it, and other checks can use it if they need the same thing.

This order is not identical to the old one. Task IDs are now sorted in config order, while `expand.grid()` varied the first column fastest. Inside each group of task ID values, the order of `output_type_id` is the same as before, and that is the only part the ascending check uses.

**`check_tbl_value_col()` requires a character `tbl`.**

**A missing task ID column is now an error.** Before, the grid supplied the missing column, and the join returned a number of rows that did not correspond to the submission. For example, removing `location` from a submission of 1,590 rows returned 450 rows and a dplyr warning about a many-to-many join.

**Derived task IDs are no longer used for matching.** They were used before, but only `NA` against `NA`: reading the config into value sets sets them to `NA`, and callers set the column to `NA` too. The result was always true. Skipping them saves one pass over the data per derived task ID. `check_tbl_value_col_ascending()` still sets them to `NA`, because it groups by these columns.

## Other changes in this PR

- `validate_derived_task_ids()` has a new `call` argument, so its warning names the function the user called.
- `_benchmark/check-calls.R` now gives `check_tbl_value_col` the character table, like `validate_model_data()`. With the typed table the check still works, but `match()` converts a `Date` column once per modeling task: 3.81s instead of 0.68s at size M with 7 modeling tasks.
- New `HUBVALIDATIONS_BENCHMARK_CHECKS` variable, to run only named checks.

## Tests

`helper-match-tbl.R` keeps `match_via_grid()`, which is the old implementation. The tests check that the new code puts the same rows in the same modeling tasks, for all 16 hubs that the test suite has a submission for. These cover schema versions v3 to v6, samples, derived task IDs, and rounds with several modeling tasks. Each hub is tested for every output type, and for both values of `subset_to_tbl_cols`.

The order is tested separately, using a config where the task IDs vary independently.

The test file runs in 5.4 seconds.

## Not in this PR

`check_tbl_values()` (#356), `check_tbl_values_required()` (#357) and the sample checks (#368) still build the grid. This is why they are unchanged in the table. `submission_tmpl()` needs the full grid, because its purpose is to generate every valid combination.


R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+
+  # Narrow to the rows the modeling task accepts, one task ID column at a time.
+  # `tbl_chr` is character and the config holds values as JSON gave them,
+  # numbers as numbers, so `%in%` coerces the config side to character to
+  # compare. It never touches the data, which is the expensive side. It also
+  # matches `NA` to `NA`, as the join it replaces did, which is how a task ID
+  # the modeling task does not use is matched.
+  keep <- !is.na(output_type_id_pos)
+  for (task_id in match_task_ids) {
+    keep <- keep & tbl_chr[[task_id]] %in% mt[["task_ids"]][[task_id]]
```
annakrystalli, line 145: This is where the type handling happens, and it is easy to miss.

`%in%` is `match(x, table, nomatch = 0L) > 0L`. `match()` converts both of its arguments to a common type, and that type is character when either side is character.

`tbl_chr` is already character, so only the config side is converted. The config holds a handful of values per column and the data holds millions of rows, so the conversion cost stays small no matter how large the submission is. Converting the data instead would be the expensive direction.

This is the reason `tbl` must be character, and why `all_character` is no longer needed.

micokoch: From Claude:
QUESTION (non-blocking): this path lets %in%/match() coerce the config side with
R's as.character(), while the checks still on the grid path (check_tbl_values,
check_tbl_values_required) coerce via hubData::coerce_to_character() (see
expand_model_out_grid.R:532). For a numeric task-ID/output_type_id value that R
renders in scientific notation (e.g. 100000 -> "1e+05" vs "100000"), could the
two paths assign a row differently — so check_tbl_value_col silently skips a row
that check_tbl_values still validates? Realistically near-zero exposure (task IDs
are dates/locations/small ints; sample IDs are set to NA before matching) and
hub-num covers ordinary numerics. Since both paths now share
extract_round_property_values() upstream, the only divergence is this final
coercion step — worth a one-line confirmation that the two render numbers
identically, or a fixture with such a value.

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+  mt,
+  match_task_ids,
+  output_type_id_pos
+) {
+  output_type_col <- tbl_chr[[hubUtils::std_colnames[["output_type"]]]]
+  task_ids <- mt[["task_ids"]]
+  # Only `output_type_id_pos` is worked out already, because matching needed it.
+  # The rest are wanted for ordering alone, so they are worked out here and only
+  # for the rows being ordered. Derived task IDs are left out, as they are for
+  # matching: every row holds `NA` there, so they cannot separate any two rows.
```
micokoch: From Claude:
NIT (comment accuracy): "every row holds NA there" is no longer generally true
now that derived task IDs are returned holding their real values — it's only true
in the one caller that blanks them (check_tbl_value_col_ascending). The exclusion
is still correct; the real reason is functional dependence (derived = f(matched
task IDs)), so it can't separate two otherwise-identical rows. Suggest rewording
to that.

R/check_tbl_value_col_ascending.R
micokoch: From Claude:
NIT (now redundant): blanking derived task IDs to NA_character_ is no longer
needed — match_tbl_to_model_task() ignores derived task IDs for matching now, so
this line can go. Harmless either way (they're constant within a group), just
dead weight.

_benchmark/prof-summary.md
```diff
@@ -135,3 +162,44 @@ when it is read with real column types.
...
+approach that could have cost more than it saved, and it does not.
+
+Nothing else moved, which is the point: `check_tbl_values` and
+`check_tbl_values_required` still build the grid, and the sample checks still work out
+model tasks their own way. A whole file at size L takes 1,771 s against a baseline of
+1,769 s, because `check_tbl_values_required` is 91% of that run and #357 is what
+addresses it.
+
+Handed the typed submission instead, `check_tbl_value_col` still works but pays
+`match()` a `Date` conversion per model task: 3.81 s against 0.68 s at size M with
```
micokoch: From Claude:
NIT (prose vs data): the typed-vs-char illustration cites "3.81 s against 0.68 s",
but the committed CSV row for that cell is 3.77 s / 0.80 s. Same story (typed
~4–5× slower because match() re-renders the Date column once per modeling task),
but the 0.68 doesn't match the committed 0.80 — reconcile so a careful reader [or computer agent 😅]
isn't puzzled.

tests/testthat/test-match_tbl_to_model_task.R
```diff
@@ -18,4 +18,145 @@ test_that("match_tbl_to_model_task works", {
...
+            output_types = output_type,
+            derived_task_ids = derived_task_ids,
+            subset_to_tbl_cols = subset_to_tbl_cols
+          )),
+          info = paste(basename(hub_path), output_type, subset_to_tbl_cols)
+        )
+      }
+    }
+  }
+})
```
micokoch, line 74: From Claude:
Nice — asserting the new partition equals match_via_grid() (the literal old path)
across 16 fixtures spanning v3–v6, samples, derived IDs, multi-task rounds, and
both hub-num/hub-chr is the right way to lock this refactor down.

R/expand_model_out_grid.R
```diff
@@ -287,6 +277,56 @@ expand_model_out_grid <- function(
...
+#' round config,
+#' set derived and unused task IDs to `NA`, fix the round ID column to the round being
+#' built, and collapse each property to the values it allows.
+#'
+#' @inheritParams expand_model_out_grid
+#'
+#' @returns A list of two, `task_ids` and `output_type`, each with one element per
+#' modeling task in the round.
+#' @noRd
+extract_round_property_values <- function(
```
micokoch, line 297: From Claude:
Good call extracting extract_round_property_values() as the shared source for
both the grid and the value-set path — this is what makes "the two can't drift
apart" real, and it's what shrinks the coercion question above to just the final
cast.

micokoch (approved):
This looks great. I went through it carefully and was able to follow the narrative and logic. I had a couple of questions/comments, but I asked Claude, who explained them and they made sense. So, there are no special comments from me, but here are Claude's close reading comments:

Gave this a proper read (algorithm, the four breaking changes, both consumer
checks, the grid refactor, and the benchmark write-up) — it's solid, approving.
The core is equivalent to the old grid-join by construction: %in%-per-column
tests the same set membership without materialising the Cartesian product. I
traced the subtle bits and they hold:
- Unused/omitted task IDs become NA, and NA %in% NA is TRUE, so a modeling task
  only accepts rows that are NA there — exactly what the grid's NA column did.
- Point-estimate output types (output_type_id = NA) match via match(NA, NA).
- Derived task IDs are correctly excluded from both matching and ordering (safe:
  a derived value is a function of the matched task IDs, so it can't break a tie
  they didn't).
- The ascending check stays correct: check_values_ascending() re-groups with
  group_by(), which preserves within-group order, so even though the new sort
  makes task-ID groups non-contiguous, each group is still output_type_id-
  ascending.
The equivalence test against match_via_grid() across all 16 fixtures is exactly
the right way to prove a refactor like this, and lifting the config-reading into
a shared extract_round_property_values() means the grid and set paths can't drift
apart. Perf numbers check out against peak-results.csv, and G3 is now actually
measured (it was defined-but-unmeasured in #360).
One question (non-blocking) on coercion parity, and a few small nits below.

_benchmark/prof-summary.md
```diff
@@ -135,3 +162,44 @@ when it is read with real column types.
...
+Three things worth reading off it.
+
+**The gaps between G1, G2 and G3 closed, rather than narrowing.** The expectation was
+that they would narrow but not close, because a config permitting more values still
+means more values to test each column against. That is true and it turns out not to
+matter. The config is read once into a lookup per column, and the per-row cost does not
+touch it again, so what grows is the lookup rather than the work. G1 to G3 multiplies
+the combinations by 6.7 while the longest single value list only goes from 365 to 730.
+
+**Both now sit at the floor.** Loading the package and reading the submission is
```
LucieContamin: I don't know what "both now sit at the floor" means here.

_benchmark/prof-summary.md
```diff
@@ -135,3 +162,44 @@ when it is read with real column types.
...
+
+Nothing else moved, which is the point: `check_tbl_values` and
+`check_tbl_values_required` still build the grid, and the sample checks still work out
+model tasks their own way. A whole file at size L takes 1,771 s against a baseline of
+1,769 s, because `check_tbl_values_required` is 91% of that run and #357 is what
+addresses it.
+
+Handed the typed submission instead, `check_tbl_value_col` still works but pays
+`match()` a `Date` conversion per model task: 3.81 s against 0.68 s at size M with
+seven.
```
LucieContamin: I don't understand that paragraph, sorry

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+#' @param derived_task_ids Character vector of derived task ID names, or `NULL`
+#' for none. A derived task ID's value is worked out from other task IDs, and
+#' those are matched on, so it adds nothing to deciding where a row belongs.
+#' These columns are not matched on and come back holding whatever they held.
+#' @param output_types Character vector of output type names to consider, or
+#' `NULL` for every output type the round defines. Checks validate one output
+#' type at a time, so this is usually a single name: a modeling task offering
+#' several will only be asked about that one, and rows of any other output type
+#' match nothing. A modeling task offering none of the requested output types
+#' gets `NULL`.
```
LucieContamin: I am not sure I understand the documentation here. `NULL` check every output types the round defines and is the default. But then you say that the checks validate one output type at a time, so it's usually a single name. It sounds a little bit in contradiction to me as a model task can have multiple output type. And the last sentence says that a modeling task with none of the requested output types gets `NULL`, does that means it returns `NULL` or the parameter should be `NULL` for no output type (in this case it contradicts the first line).

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+#' @param subset_to_tbl_cols Logical. Which columns to return. `TRUE` gives the
+#' columns of `tbl_chr`, in their own order, without `value`. `FALSE` gives the
+#' columns the modeling task itself defines: its task IDs in config order, then
+#' `output_type` and `output_type_id`, with `value` last when `tbl_chr` has one.
```
LucieContamin:
```suggestion
#' @param subset_to_tbl_cols Logical. Which columns to return. `TRUE` (default) gives the
#' columns of `tbl_chr`, in their own order, without `value`. `FALSE` gives the
#' columns the modeling task itself defines: its task IDs in config order, then
#' `output_type` and `output_type_id`, with `value` last when `tbl_chr` has one.
```
might be good to inform what is the default behavior

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+#' @param order_by_config Logical. Whether to sort each modeling task's rows by
+#' where each value sits in the config's list for its column, rather than
+#' leaving them in the order they were submitted in. Sorted on `output_type`, so
+#' rows of one output type sit together, then `output_type_id` so they ascend
+#' within each, then the task IDs to break ties. Positions are compared, not the
+#' values themselves, so a column whose values do not sort meaningfully still
+#' comes back in the order the config gives them. See `order_by_config()`.
```
LucieContamin:
```suggestion
#' @param order_by_config Logical. Whether to sort each modeling task's rows by
#' where each value sits in the config's list for its column, rather than
#' leaving them in the order they were submitted in, `FALSE` by default. Sorted on `output_type`, so
#' rows of one output type sit together, then `output_type_id` so they ascend
#' within each, then the task IDs to break ties. Positions are compared, not the
#' values themselves, so a column whose values do not sort meaningfully still
#' comes back in the order the config gives them. See `order_by_config()`.
```

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+    derived_task_ids = derived_task_ids,
+    call = rlang::caller_env()
+  )
+
+  # `get_config_mt_value_sets()` gives every modeling task an entry for every
+  # task ID in the round, so the first one already lists them all. Those, plus
+  # the two output type columns, are the columns `tbl_chr` must have to be
+  # matched.
+  match_cols <- c(
+    names(value_sets[[1L]][["task_ids"]]),
```
LucieContamin: What happen if value sets has multiple element and the first one is having one of the null column (for example horizon) and it got removed in the `value_sets` list creation? It might cause issue, no? would it not be better to extract all the task id columns for all element?

LucieContamin: What happen if value sets has multiple element and the first one is having one of the null column (for example horizon) and it got removed in the `value_sets` list creation? It might cause issue, no? would it not be better to extract all the task id columns for all element?

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+  tbl_chr,
+  mt,
+  derived_task_ids,
+  subset_to_tbl_cols,
+  order_by_config
+) {
+  # A derived task ID's value is worked out from other task IDs, and those are
+  # matched on, so it can only agree with them about where a row belongs. It is
+  # left out, which saves a pass over the data each.
+  match_task_ids <- setdiff(names(mt[["task_ids"]]), derived_task_ids)
```
LucieContamin: I understand globally the idea here to remove the derived task ids from the match, but the documentation is really not clear to me.

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+    return(NULL)
+  }
+  output_type_col <- tbl_chr[[hubUtils::std_colnames[["output_type"]]]]
+  output_type_id_col <- tbl_chr[[hubUtils::std_colnames[["output_type_id"]]]]
+
+  # Collect the positions of `output_type_id`s in relation to the order
+  # specified in the config for each output type. Each output type has its own
+  # order, which is why this is a loop rather than one lookup. The `NA`s left
+  # behind drop rows this modeling task does not accept, and the positions are
+  # what `order_by_config` sorts on.
```
LucieContamin: I am not clear on what the last phrase means, sorry

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+      mt[["output_type_ids"]][[output_type]]
+    )
+  }
+
+  # Narrow to the rows the modeling task accepts, one task ID column at a time.
+  # `tbl_chr` is character and the config holds values as JSON gave them,
+  # numbers as numbers, so `%in%` coerces the config side to character to
+  # compare. It never touches the data, which is the expensive side. It also
+  # matches `NA` to `NA`, as the join it replaces did, which is how a task ID
+  # the modeling task does not use is matched.
```
LucieContamin: Not sure I understand the "which is how a task ID the modeling task does not use is matched." Especially if the task id was removed from the `mt` object.

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+# type first, so rows of one output type are together, then output type ID, so
+# they ascend within each, then the task IDs to break ties. Positions are
+# compared rather than the values themselves, because output type IDs do not
+# always sort meaningfully. Sorting is stable, so rows sharing a position keep
+# the order they were submitted in.
+#
+# Belongs here rather than in the caller because working out where each value
+# sits in the config is already part of assigning a row. A caller sorting for
+# itself would have to read the config again to recover positions this function
+# is handed for free.
```
LucieContamin: Not sure I understand that paragraph, sorry

R/assign_tbl_to_model_task.R
LucieContamin: I think I understand the logic here but this paragraph was difficult for me, can we simplify or rewrite it (ignore if it's only me)

R/assign_tbl_to_model_task.R
```diff
@@ -0,0 +1,208 @@
...
+    list(
+      match(output_type_col[rows], names(mt[["output_type_ids"]])),
+      output_type_id_pos[rows]
+    ),
+    purrr::map(
+      match_task_ids,
+      \(task_id) match(tbl_chr[[task_id]][rows], task_ids[[task_id]])
+    )
+  )
+  rows[do.call(order, c(keys, list(method = "radix")))]
```
LucieContamin, line 223: I am not sure I understand how the order works here with a list of row order as input. Does it overwrite itself for each list element?

R/match_tbl_to_model_task.R
```diff
@@ -1,10 +1,30 @@
...
+#' Column types must **all be character**: values are matched against the
+#' config's with `%in%`, which converts the config side to character to compare,
+#' so the data is never converted and never has to be. That conversion uses R's
+#' own rendering of a number, so a config value R prints in scientific notation
+#' does not match its plain form in the data. Every task ID column the round
+#' defines must be present.
+#' @param derived_task_ids Character vector of derived task ID names, or `NULL`
+#' for none. A derived task ID's value is worked out from other task IDs, and
+#' those are matched on, so it adds nothing to deciding where a row belongs.
+#' These columns are not matched on and come back holding whatever they held.
```
LucieContamin: I find the phrasing here a little bit confusing. (feel free to ignore if it's just me)

R/match_tbl_to_model_task.R
```diff
@@ -1,10 +1,30 @@
...
+#' @param order_by_config Logical. Whether to sort each modeling task's rows by
+#' where each of their values sits in the config, rather than leaving them in the
+#' order they were submitted in. Rows are sorted on `output_type` first, so rows of
+#' one output type sit together, then on `output_type_id` so they ascend within
+#' each, then on the task IDs to break ties. What is compared is each value's
+#' position in the config's list for its column, not the value itself, so a column
+#' whose values do not sort meaningfully still comes back in the order the config
+#' gives them. That is what a check reading values in sequence needs, such as the
```
LucieContamin: What does meaningfully means here?

R/match_tbl_to_model_task.R
```diff
@@ -1,10 +1,30 @@
...
+#' non-descending check on `quantile` and `cdf` values.
```
LucieContamin: The last phrase is not clear to me, sorry

R/expand_model_out_grid.R
```diff
@@ -287,6 +277,56 @@ expand_model_out_grid <- function(
   )
 }
 
+#' Read the values each modeling task lists for each of its columns
+#'
+#' The preparation shared by [expand_model_out_grid()], which expands these values into
+#' a grid, and `get_config_mt_value_sets()`, which keeps them as sets: read the
+#' round config,
```
LucieContamin: It took me a minute to understand the documentation, are you talking on how to prepared the grid?

R/utils-get_config.R
```diff
@@ -26,3 +26,90 @@ get_config_derived_task_ids <- function(config_tasks, round_id = NULL) {
...
+#'     $pmf      "low" "moderate" "high" ...
+#' ```
+#'
+#' So a row belongs to that modeling task when its `target` is that one string,
+#' its `horizon` is one of those four numbers, and so on. That is what matching
+#' needs: within a modeling task the valid combinations are the Cartesian
+#' product of these lists, so testing a row against them answers the same
+#' question as matching it against the expanded product, without building the
+#' product.
+#'
```
LucieContamin: I am not sure its the best place for that documentation as the function is not about the matching. I understand we use that function for the matching, but if we want to use it for something else, it might become a little bit confusing.

R/utils-get_config.R
```diff
@@ -26,3 +26,90 @@ get_config_derived_task_ids <- function(config_tasks, round_id = NULL) {
...
+#' Three things about configs make this more than a lookup. A task ID a modeling
+#' task does not use is listed as `null`, or left out; which `output_type_id`
+#' values are valid depends on the output type, so they are keyed by it rather
```
LucieContamin: It might be good to clarify when its listed as NULL or left out as it might have impact on the expected output and steps taken after that function

R/utils-get_config.R
```diff
@@ -26,3 +26,90 @@ get_config_derived_task_ids <- function(config_tasks, round_id = NULL) {
...
+#' than pooled; and `round_id` has to be pinned to the round being read when it
```
LucieContamin: What do you mean by "keyed by it rather than pooled"?

R/utils-get_config.R
```diff
@@ -26,3 +26,90 @@ get_config_derived_task_ids <- function(config_tasks, round_id = NULL) {
...
+#' comes from a variable. `extract_round_property_values()` handles those,
+#' shared with [expand_model_out_grid()] so the two cannot drift apart.
+#'
+#' @inheritParams expand_model_out_grid
+#'
+#' @returns A list with one element per modeling task in the round, each a list
+#' of `task_ids` and `output_type_ids` as above. Values come back as the config
+#' holds them, numbers as numbers. Nothing needs coercing here because they are
```
LucieContamin: "Values come back as the config holds them, numbers as numbers" , do you mean the type of the object is respected?

R/utils-get_config.R
```diff
@@ -26,3 +26,90 @@ get_config_derived_task_ids <- function(config_tasks, round_id = NULL) {
...
+
+  round_task_ids <- hubUtils::get_round_task_id_names(config_tasks, round_id)
+
+  purrr::map2(
+    purrr::map(property_values[["task_ids"]], purrr::compact),
+    property_values[["output_type"]],
+    \(task_ids, output_type) {
+      # A modeling task that does not use a task ID lists it as null, or
+      # leaves it out altogether. A peak target has no `horizon`, for
+      # instance, and its rows carry `NA` in that column, so `NA` is the only
```
LucieContamin:
```suggestion
      # leaves it out altogether. A peak target might not have `horizon`, for
      # instance, and its rows carry `NA` in that column, so `NA` is the only
```
We can imagine a peak timing target using the horizon column

R/utils-get_config.R
```diff
@@ -26,3 +26,90 @@ get_config_derived_task_ids <- function(config_tasks, round_id = NULL) {
...
+      # value the modeling task accepts there. `null_taskids_to_na()` has
+      # already handled the ones listed as null; this handles the ones left
```
LucieContamin: Where is `null_taskids_to_na()` used?

NEWS.md
```diff
@@ -1,7 +1,18 @@
...
+
+## Other changes
+
 * A `compound_taskid_set` of `[]`, which declares that every task ID is sampled jointly, no longer errors. `check_tbl_spl_compound_taskid_set()` failed with `` `config_comp_tids` must be logical, numeric, or character, not an empty list `` and `submission_tmpl()` with `invalid subscript type 'list'`. The check now passes when a submission's samples are drawn jointly across every task ID and, when they are finer, reports that the hub expects no compound task IDs and response dependence across all task IDs (#361).
 * `get_tbl_compound_taskid_set()` no longer drops modeling tasks whose detected compound task ID set is empty, which previously made a jointly sampled modeling task indistinguishable from one with no samples at all (#361).
+* `match_tbl_to_model_task()` no longer matches on derived task ID columns, so their values are left alone and matched rows come back holding them. A derived task ID's value is worked out from other task IDs, which are matched on, so it never decided which modeling task a row belonged to. It previously had to hold `NA` for a row to match anything, which meant blanking the column beforehand (#355).
```
LucieContamin: Sorry, I am not clear on the last part of the paragraph: "so it never decided which modeling task a row belonged to. It previously had to hold `NA` for a row to match anything, which meant blanking the column beforehand (#355)."

NEWS.md
```diff
@@ -1,7 +1,18 @@
...
+* `match_tbl_to_model_task()` now works out which modeling task each row belongs to by testing each column against the values that modeling task allows, instead of building the grid of every valid value combination and joining to it. It assigns the same rows, but on a hub whose config allows a large number of combinations it is much faster and needs far less memory, as are `check_tbl_value_col()` and `check_tbl_value_col_ascending()`, which use it (#355).
```
LucieContamin:
```suggestion
* `match_tbl_to_model_task()` now works out which modeling task each row belongs to by testing each column against the values that modeling task allows, instead of building the grid of every valid value combination and joining to it. It generates the same results, but on a hub whose config allows a large number of combinations it is much faster and needs far less memory, as are `check_tbl_value_col()` and `check_tbl_value_col_ascending()`, which use it (#355).
```

LucieContamin (changes requested):
Thank you for the update, it generally looks great and happy to see that the performance was increased. 

I just wanted to check that the new match system keeps only row that are expected from the config files, the output is only used in check to test the variables types and value increase and not the content of the model output (test if its missing some specific value, like a specific target for a location, etc.). Am I correct?

Also, I request changes, because I have some question about some code, but mainly I really struggle with the documentation. It took me a long time to read all of it and try to make sense of some part of it. If it is only me, maybe ignore but I think it will be good to clarify some points, even for the internal documentation.

R/assign_tbl_to_model_task.R
annakrystalli: Good catch, and the answer was no, they did not render identically.

`%in%` used R's `as.character()`, while the grid path and the all-character copy of a submission both go through `hubData::coerce_to_character()`, which casts via arrow. Two ways those diverge:

- a large number: `as.character(1e5)` is `"1e+05"`, arrow gives `"100000"`
- a logical: `as.character(TRUE)` is `"TRUE"`, arrow gives `"true"`

Neither is reachable with the task ID types hubs actually use, so no hub was affected. But since both paths already share `extract_round_property_values()`, it was cheap to close the last gap rather than document it.

`get_config_mt_value_sets()` now casts the values through arrow as it extracts them, so the config side renders exactly as it did on the grid path. Matching is character against character, and the caveat is gone from the docs rather than reworded.

The logical case did turn out to be a genuine gap beyond this PR: csv and parquet submissions render logicals differently, so a logical task ID matches inconsistently between them. Opened as #371, not urgent since no hub uses logicals.

annakrystalli:
On your first question: yes, and none of that changes here.

Matching returns only rows whose values the config allows and drops the rest, exactly as the grid join did. Its consumers are `check_tbl_value_col()` and `check_tbl_value_col_ascending()`, neither of which asks whether anything is missing. The checks that do test for missing or unexpected content, `check_tbl_values()` and `check_tbl_values_required()`, do not use it — they still build the grid, and #356 and #357 cover them.

One thing worth separating out, because it also came up on the `output_types` argument: validating one output type at a time is not new, and not something this PR introduces. `expand_model_out_grid()` has taken an `output_types` argument for a long time, and `check_tbl_value_col_ascending()` has looped over the output types present in a submission since January 2025 (b021028). It is not confined to the checks touched here, either — `check_tbl_values()` splits the submission by `output_type`, and `check_tbl_values_required()` maps over output types deliberately, to reduce memory pressure and to keep their evaluation separate (#177). Those two are where this approach goes next, in #356 and #357, so the pattern carries straight over. A modeling task can certainly define several output types; the checks just ask about one at a time.

On the documentation: that was fair, and it was not only you. I have been through every comment; 6cafdd9 has the rewrites. Thanks for being specific about which passages did not land, it made them easy to find.

_benchmark/prof-summary.md
annakrystalli: Fixed in 6cafdd9 — dropped the "floor" idiom here and in the three other places it appeared.

_benchmark/prof-summary.md
annakrystalli: Fixed in 6cafdd9. The figures were also wrong against the committed CSV; corrected to 3.77 s / 0.80 s.

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9. You were right that it contradicted itself — it never said the argument subsets the data as well as matching it, and the `NULL` return is now left to `@returns`.

R/assign_tbl_to_model_task.R
annakrystalli: Applied in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: Applied in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: No bug, but the code was making you work too hard to see that, which is why I've changed it anyway.

`get_config_mt_value_sets()` gives every modeling task an entry for every task ID the round defines, including ones it does not use: `omitted <- setdiff(round_task_ids, names(task_ids)); task_ids[omitted] <- list(NA)`. So the first element does list them all. I checked both of the cases you describe — `horizon` declared `null` in mt1, and `horizon` left out of mt1 entirely — and it is present with value `NA` either way.

But that relies on an invariant set up in a different file, which you cannot see from this line. `match_cols` now comes from `hubUtils::get_round_task_id_names(config_tasks, round_id)`, which says what is meant directly. 6cafdd9

annakrystalli: No bug, but the code was making you work too hard to see that, which is why I've changed it anyway.

`get_config_mt_value_sets()` gives every modeling task an entry for every task ID the round defines, including ones it does not use: `omitted <- setdiff(round_task_ids, names(task_ids)); task_ids[omitted] <- list(NA)`. So the first element does list them all. I checked both of the cases you describe — `horizon` declared `null` in mt1, and `horizon` left out of mt1 entirely — and it is present with value `NA` either way.

But that relies on an invariant set up in a different file, which you cannot see from this line. `match_cols` now comes from `hubUtils::get_round_task_id_names(config_tasks, round_id)`, which says what is meant directly. 6cafdd9

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/assign_tbl_to_model_task.R
annakrystalli: `order()` takes any number of vectors and uses them as successive tie-breakers: the first decides, the second only separates rows the first leaves equal, and so on. Nothing overwrites anything. `do.call()` is there because the keys are built up in a list and `order()` needs them as separate arguments.

Added a comment saying so, since it clearly was not obvious. 6cafdd9

R/match_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/match_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/match_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9.

R/expand_model_out_grid.R
annakrystalli: Fixed in 6cafdd9.

R/utils-get_config.R
annakrystalli: Fixed in 6cafdd9 — agreed, it was describing what the caller does with the output rather than what the function does.

R/utils-get_config.R
annakrystalli: Fixed in 6cafdd9.

R/utils-get_config.R
annakrystalli: Fixed in 6cafdd9.

R/utils-get_config.R
annakrystalli: That sentence was describing something that turned out to be a bug, so it is gone rather than clarified. The values were left as the config held them and `%in%` converted them on every comparison, using R's rendering rather than the arrow one the rest of the pipeline uses. They are now converted once, through arrow, when they are extracted — so the answer to your question is no, the type is not preserved, they come back as character. d7e788b, and see Alvaro's thread above.

R/utils-get_config.R
annakrystalli: Applied in 6cafdd9.

R/utils-get_config.R
annakrystalli: `extract_round_property_values()`, in `expand_model_out_grid.R`. The comment now names it. Fixed in 6cafdd9.

NEWS.md
annakrystalli: Fixed in 6cafdd9 — that was implementation mechanics which should not have been in NEWS at all.

NEWS.md
annakrystalli: Applied in 6cafdd9, close to your wording.

R/assign_tbl_to_model_task.R
annakrystalli: Fixed in 6cafdd9 — correct, that stopped being true once derived task IDs came back with their values.

R/check_tbl_value_col_ascending.R
annakrystalli: Right that matching no longer needs it, but removing it outright breaks `check_tbl_value_col_ascending()` — there's a regression test for exactly this, #189, which corrupts `target_end_date` and expects `check_success`.

The reason is that `check_values_ascending()` groups by every non-standard column, derived ones included. For a well-formed file that changes nothing, since a derived value is constant within a group. For a file whose derived values are wrong it splits the group, and a descending value inside it stops being reported. On hub-177, one group of 23 quantile rows with the values deliberately reversed:

```
derived value consistent    -> blanked: 1 problem found   not blanked: 1 problem found
derived value inconsistent  -> blanked: 1 problem found   not blanked: 0, error hidden
```

That state does reach this check: `check_tbl_derived_task_id_vals()` returns a `check_failure` rather than an error, so there is no early return between it and here.

But you were pointing at something real. Ignoring derived task IDs in matching while mutating the data to achieve the same thing in grouping is one policy implemented two ways. `check_values_ascending()` now leaves derived task IDs out of `group_cols` directly, the same way `assign_mt_row_idx()` leaves them out of matching. The blanking is gone, the submission is no longer mutated, and the error table no longer carries an all-`NA` column per derived task ID. 55b4637

_benchmark/prof-summary.md
annakrystalli: Fixed in 6cafdd9. I also re-measured to be sure, and the CSV figures are right: 3.77 s / 0.80 s.

tests/testthat/test-match_tbl_to_model_task.R
annakrystalli: Thanks — it's the part I'd have wanted to see too. #368 does the same for the sample checks.

R/expand_model_out_grid.R
annakrystalli: Thanks. It has since earned its keep: it is now also the single place config values are converted to character, which closes the coercion gap you asked about above. d7e788b

_benchmark/prof-summary.md
```diff
@@ -97,15 +124,16 @@ time for one check at one size.
...
+really use, not a measurement of it.
 
 The last two rows are the control. Neither check has to work out which model task a row
 belongs to, so neither is affected by the size of the grid: both sit at ~200-290 MB and
 under a second across a grid that grew nearly sevenfold.
 
 What to expect of the other rows once #355-#357 land: they should **drop**, and the gaps
-between the G columns should **narrow**. Not to the controls' floor, though. A config
+between the G columns should **narrow**, though not all the way down to the controls. A
+config
```
LucieContamin: Very minor note, `config` is alone on the line.

_benchmark/prof-summary.md
```diff
@@ -135,3 +163,46 @@ when it is read with real column types.
...
+Three things worth reading off it.
+
+**The gaps between G1, G2 and G3 closed, rather than narrowing.** The expectation was
+that they would narrow but not close, because a config permitting more values still
+means more values to test each column against. That is true and it turns out not to
+matter. The config is read once into a lookup per column, and the per-row cost does not
+touch it again, so what grows is the lookup rather than the work. G1 to G3 multiplies
+the combinations by 6.7 while the longest single value list only goes from 365 to 730.
+
+**Both now cost almost nothing beyond the setup.** Loading the package and reading
```
LucieContamin: Minor suggestion: the "Both" is not clear until we read the second sentence.

R/match_tbl_to_model_task.R
```diff
@@ -1,10 +1,34 @@
...
+#' skipped, and returned unchanged.
+#' @param order_by_config Logical. How to order each modeling task's rows.
+#' `FALSE`, the default, leaves them in the order they were submitted in. `TRUE`
+#' sorts them into the order the config lists their values in: on `output_type`
+#' first, so rows of one output type sit together, then on `output_type_id`, so
+#' they ascend within each, then on the task IDs to break ties.
+#'
+#' What is sorted on is each value's position in the config, not the value
+#' itself. `pmf` categories show why that matters: `"low"`, `"moderate"` and
+#' `"high"` have no useful alphabetical order, but the config lists them in the
```
LucieContamin, line 25:
```suggestion
#' itself. Example of `pmf` categories show why that matters: `"low"`, `"moderate"` and
#' `"high"` have no useful alphabetical order, but the config lists them in the
```

Minor suggestion

LucieContamin (approved):
Thank you very much for the update and the additional information. It all looks good to me! I made some minor suggestion, feel free to ignore.

PR #373, opened 2026-08-17
Assign sample rows to modeling tasks without expanding the grid

### LINKED ISSUE #368: Grid-free sample checks: assign sample rows by set membership

annakrystalli:
Resolves #368.

The sample checks worked out which modeling task each sample row belongs to by expanding the grid of every valid value combination for that modeling task and joining the data to it. They now test each column against the values that modeling task allows, which is what #355 does for the other checks. Nothing larger than the submission is built.

`spl_hash_tbl()` also used to read the compound index off the grid. In a template the sample ID *is* the compound index: `expand_model_out_grid(include_sample_ids = TRUE)` gives one ID per distinct combination of compound task ID values, and `spl_hash_tbl()` renamed that column to `compound_idx` and used it as it stood. The index is now worked out from the config's value lists instead, as the position of a row's combination among the combinations the modeling task allows. It appears in check messages and is the sample ID that `submission_tmpl()` writes, so the numbering had to stay the same. `add_mt_sample_idx()` still numbers by expanding the grid, because `submission_tmpl()` has to build the grid anyway, so changes to either function need syncing with the other.

**Sample `output_type_id` values are now returned as submitted.** They are identifiers the submitter chose, so the config lists no ids to compare them against and the comparison could never reject a row. Sample structure is validated downstream by the `check_tbl_spl_*` checks, so nothing is lost by matching not looking at the ids. `match_tbl_to_model_task()` used to set the column to `NA` before matching, so that a real id could not fail it, which lost the ids from the returned tables. `match_output_type_ids()` now skips the comparison for samples instead, and the blanking is gone. This is the same argument that stopped derived task IDs being blanked in #355.

`assign_mt_rows()` is split into named stages: `match_output_type_ids()` and `narrow_by_task_id_values()`, with `subset_mt_tbl()` building the result.

Samples have no config order to sort on, so with `order_by_config = TRUE` they tie on the `output_type_id` key and are ordered by their task ID values alone, which does not keep a sample's rows together. #376 covers whether to group them.

## Performance

At G3, which is 65,000 submitted rows against a config permitting 77.7 million valid combinations, the five sample checks go from 13 to 14 GB and 29 to 78 seconds each, down to under 300 MB and under half a second. `_benchmark/prof-summary.md` has the full table. At size M they barely move, because what they spend their time on there is splitting rows one sample at a time, not the grid.

## Testing

`tests/testthat/test-spl-assign.R` keeps the old grid implementations in `helper-spl-assign.R` and compares the two on six sample fixtures, for both configured and detected compound task ID sets. There are also tests for a modeling task with no submitted samples, a `compound_taskid_set` written in a different order from the config, an unknown task ID name, a set of the wrong length, and an index large enough that `as.character()` would write it in scientific notation.

Two checks that `add_mt_sample_idx()` used to perform on the grid path are now shared, so they still run when the grid is not built: the `compound_taskid_set` length, and its task ID names. Before this, a wrong length gave `subscript out of bounds` and an unknown name was dropped silently.

## Merge order

This targets `ak/grid-free-assignment/355` (#369). Once that merges, GitHub will retarget this to `main`.


tests/testthat/_snaps/spl-assign.md
```diff
@@ -0,0 +1,20 @@
...
+      i The `compound_taskid_set` must be a subset of "reference_date", "target", "horizon", and "location".
+
+# the sample checks still validate compound_taskid_set length without building the grid
+
+    Code
+      check_tbl_spl_n(tbl, "2022-10-22", "team-model/2022-10-22-team-model.csv",
+        hub_path, compound_taskid_set = list(c("target", "location")))
+    Condition
+      Error in `check_tbl_spl_n()`:
+      x The length of `compound_taskid_set` (1) must match the number of modeling tasks (2) in the round.
```
LucieContamin: I did not notice before, but I found the message here a bit confusing. I read it as the length of the`compound_taskid_set` (meaning the column defining the `compound_taskid_set` for example "target", "horizon", etc.) must match the number the modeling tasks in the round (meaning number of different modeling tasks possible). 
Or I thought the `check_tbl_spl_n` checks that the number of trajectories (or samples) submitted are in the range defined in the tasks.json. For example, a modeling task accept between 100 and 300 trajectories. Did I miss something?

LucieContamin (commented):
Thank you for the update, it all looks good! I just have a comment on an error message.

tests/testthat/_snaps/spl-assign.md
annakrystalli: Good catch, that wording was ambiguous. `compound_taskid_set` is a list with one element per modeling task, so "length" meant the number of elements in the list, not the number of task IDs inside one. Fixed in 20c5290:

```
x `compound_taskid_set` must have one element for each modeling task in the round.
i The round has 2 modeling tasks but `compound_taskid_set` has 1.
```

On your second question you have it right. `check_tbl_spl_n()` does check that the number of samples per compound idx is consistent and falls within the `n_min` and `n_max` the config sets. This test passes it a deliberately wrong `compound_taskid_set`, which aborts before any of that runs. Inside `validate_model_data()` that abort surfaces as an `<error/check_exec_error>`, so the snapshot pins the exec error a user gets from a bad argument rather than anything about sample counts.

LucieContamin (approved):
Thank you for the update and additional information! It looks good to me

